PR #9, opened 2024-08-09
Send the "started" event on agent start

foca:
This mostly just sets up the infra to report events, and implements the started event.

Missing is sending packages/stack in the agent info, and some tests for the startup procedure / Runner object.

### DIFF
--- a/lib/aikido/agent/runner.rb
+++ b/lib/aikido/agent/runner.rb
@@ -59,7 +59,6 @@ module Aikido::Agent
     end
 
     private def reporting_pool
-      # TODO: Should we use a pool with more than one thread for parallelism?
       @reporting_pool ||= Concurrent::SingleThreadExecutor.new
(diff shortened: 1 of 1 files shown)


lib/aikido/agent/runner.rb
```diff
@@ -1,35 +1,66 @@
...
       end
     end
 
     def stop!
       @poll_for_setting_updates&.kill
+      @reporting_pool&.shutdown and @reporting_pool&.wait_for_termination(30)
+    end
+
+    private def reporting_pool
+      # TODO: Should we use a pool with more than one thread for parallelism?
```
willem-delbare: since it might be open source, let's park todo's in the jira?

foca: https://aikido.atlassian.net/browse/AIK-3332

PR #11, opened 2024-08-14
Properly implement sinks and start stat gathering

foca:
This introduces a "proper" abstraction for sinks, that are able to run all the scanners declared in a given "interceptor" (aka the module we use to hijack libraries). The sinks gather scan-specific information in a Scan object, that tracks stats like duration, and which can then be funneled somewhere else.

This keeps the actual interception logic pretty thin—all we need to do before calling super is tell the sink to scan the code, and pass any metadata we want the scanners to receive.

This also lays the foundation to notify the server of an infraction, by defining Attack objects that contain all the information captured from the actual offending call, and which can then be used to either populate the event that gets sent to the server, or generate a nice error message to show in the logs (and/or in the exception that blocks a request).

Pending (for the next PRs, because this one is getting a little big):

* Actually gather the statistics (for now we're just ignoring that).
* Capture the current "route" and track that in the request object.
* Implement a bunch of JSON serializations so that the attack events have the expected payload.
* Differentiate the user inputs based on source (currently we are just mixing path, query and body params into a single bag, which makes it difficult to report the `source` in the event payload. 

With that we should be basically on par with the JS implementation when reporting attacks, and with the routes and statistics gathering, we should be ready to implement heartbeats, which should be enough to start putting this in "actual" apps and start seeing some (hopefully useful) data.

### DIFF
--- a/test/aikido/firewall/attack_test.rb
+++ b/test/aikido/firewall/attack_test.rb
@@ -34,7 +34,7 @@ module Aikido::Firewall
       TXT
     end
 
-    test "correclty identifies the MySQL dialect in the log message" do
+    test "correctly identifies the MySQL dialect in the log message" do
       dialect = Aikido::Firewall::Vulnerabilities::SQLInjection[:mysql]
       attack = Aikido::Firewall::Attacks::SQLInjectionAttack.new(
         query: @query, input: @input, dialect: dialect, sink: @sink, request: @request
@@ -43,7 +43,7 @@ module Aikido::Firewall
       assert_match(/in MySQL query/, attack.log_message)
     end
 
-    test "correclty identifies the PostgreSQL dialect in the log message" do
+    test "correctly identifies the PostgreSQL dialect in the log message" do
       dialect = Aikido::Firewall::Vulnerabilities::SQLInjection[:postgresql]
       attack = Aikido::Firewall::Attacks::SQLInjectionAttack.new(
         query: @query, input: @input, dialect: dialect, sink: @sink, request: @request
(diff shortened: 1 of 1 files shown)


test/aikido/firewall/attack_test.rb
```diff
@@ -0,0 +1,76 @@
...
+      attack = Aikido::Firewall::Attacks::SQLInjectionAttack.new(
+        query: @query, input: @input, dialect: @dialect, sink: @sink, request: @request
+      )
+
+      assert_equal <<~TXT.chomp, attack.log_message
+        SQL Injection: Malicious user input «' OR 1=1 --» detected in SQL query «SELECT * FROM users WHERE id = '' OR 1=1 --'»
+      TXT
+    end
+
+    test "correclty identifies the MySQL dialect in the log message" do
```
hansott:
```suggestion
    test "correctly identifies the MySQL dialect in the log message" do
```

test/aikido/firewall/attack_test.rb
```diff
@@ -0,0 +1,76 @@
...
+      dialect = Aikido::Firewall::Vulnerabilities::SQLInjection[:mysql]
+      attack = Aikido::Firewall::Attacks::SQLInjectionAttack.new(
+        query: @query, input: @input, dialect: dialect, sink: @sink, request: @request
+      )
+
+      assert_match(/in MySQL query/, attack.log_message)
+    end
+
+    test "correclty identifies the PostgreSQL dialect in the log message" do
```
hansott:
```suggestion
    test "correctly identifies the PostgreSQL dialect in the log message" do
```

test/aikido/firewall/attack_test.rb
```diff
@@ -0,0 +1,76 @@
...
+      dialect = Aikido::Firewall::Vulnerabilities::SQLInjection[:postgresql]
+      attack = Aikido::Firewall::Attacks::SQLInjectionAttack.new(
+        query: @query, input: @input, dialect: dialect, sink: @sink, request: @request
+      )
+
+      assert_match(/in PostgreSQL query/, attack.log_message)
+    end
+
+    test "correclty identifies the SQLite dialect in the log message" do
```
hansott:
```suggestion
    test "correctly identifies the SQLite dialect in the log message" do
```

PR #14, opened 2024-08-18
Flesh out attack JSON payloads

foca:
This adds most of the fields missing from the `attack_detected` events.

In order to determine the correct "source" and "pathToPayload" we need to change how we read parameters from the request so we can actually trace this (currently, we're just aggregating all parameters in a single bag regardless of whether they are path, query, or body params, so it's not possible to correctly inform the user).

### DIFF
--- a/lib/aikido/firewall/sinks/pg.rb
+++ b/lib/aikido/firewall/sinks/pg.rb
@@ -14,7 +14,7 @@ module Aikido::Firewall
         ].each do |method|
           module_eval <<~RUBY, __FILE__, __LINE__ + 1
             def #{method}(query, *)
-              SINK.scan(query: query, dialect: :postgresql, operation: #{method.inspect})
+              SINK.scan(query: query, dialect: :postgresql, operation: :#{method})
               super
             rescue Aikido::Firewall::SQLInjectionError
               # The pg adapter does not wrap exceptions in StatementInvalid, which
@@ -31,7 +31,7 @@ module Aikido::Firewall
         ].each do |method|
           module_eval <<~RUBY, __FILE__, __LINE__ + 1
             def #{method}(_, query, *)
-              SINK.scan(query: query, dialect: :postgresql, operation: #{method.inspect})
+              SINK.scan(query: query, dialect: :postgresql, operation: :#{method})
               super
             rescue Aikido::Firewall::SQLInjectionError
               # The pg adapter does not wrap exceptions in StatementInvalid, which


lib/aikido/firewall/attack.rb
```diff
@@ -48,8 +59,15 @@ def log_message
...
-        # TODO: Actually implement this.
-        {kind: "sql_injection"}
+        {
+          kind: "sql_injection",
+          blocked: blocked?,
+          payload: @input,
+          metadata: {sql: @query},
+          operation: @operation,
+          source: nil,
+          pathToPayload: nil
```
hansott, line 69: We don't have the path available yet? When extracting the user input strings?

lib/aikido/firewall/sinks/pg.rb
```diff
@@ -1,18 +1,20 @@
...
       module Extensions
-        %w[
+        %i[
           send_query exec sync_exec async_exec
           send_query_params exec_params sync_exec_params async_exec_params
         ].each do |method|
           module_eval <<~RUBY, __FILE__, __LINE__ + 1
             def #{method}(query, *)
-              SINK.scan(query: query, dialect: :postgresql)
+              SINK.scan(query: query, dialect: :postgresql, operation: #{method.inspect})
```
hansott: doesn't this include a colon?

lib/aikido/firewall/sinks/pg.rb
```diff
@@ -24,12 +26,12 @@ def #{method}(query, *)
...
         end
 
-        %w[
+        %i[
           send_prepare prepare async_prepare sync_prepare
         ].each do |method|
           module_eval <<~RUBY, __FILE__, __LINE__ + 1
             def #{method}(_, query, *)
-              SINK.scan(query: query, dialect: :postgresql)
+              SINK.scan(query: query, dialect: :postgresql, operation: #{method.inspect})
```
hansott: doesn't this include a colon?

lib/aikido/firewall/sinks/pg.rb
foca: This is pretty opaque. Because this is now an array of symbols (`%i[ ... ]`), `Symbol#inspect` will output the colon. If it's more clear, happy to just change this to `:#{method}` and rely on the string representation of the symbol removing the colon.

``` ruby
:foo.inspect #=> ":foo"
:foo.to_s #=> "foo"
```

It's the same either way, so whatever makes it easier to read is good.

foca:
```suggestion
              SINK.scan(query: query, dialect: :postgresql, operation: :#{method})
```

lib/aikido/firewall/sinks/pg.rb
foca:
```suggestion
              SINK.scan(query: query, dialect: :postgresql, operation: :#{method})
```

lib/aikido/firewall/attack.rb
foca: This is the next PR which I'll open shortly :)

PR #20, opened 2024-08-28
Fix attack reporting on GET requests

foca:
(Technically, requests with no request body)

Puma's NullIO does not implement the full IO interface, which means that for requests with an empty body, we were getting an object we couldn't fully read.

I [opened a PR](https://github.com/puma/puma/pull/3468) to fix this upstream but we still need to support all versions of puma, so here we are.

### DIFF
--- a/lib/aikido/agent/request.rb
+++ b/lib/aikido/agent/request.rb
@@ -28,10 +28,19 @@ module Aikido::Agent
         }
     end
 
-    # @return [String] the request body, up to a maximum of 16KiB. If the
-    #   underlying IO object had been partially (or fully) read before,
-    #   this will restore the previous cursor position after reading it if
-    #   possible, or left it rewund if not.
+    # @api private
+    #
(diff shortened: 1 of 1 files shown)


lib/aikido/agent/request.rb
```diff
@@ -30,18 +30,19 @@ def normalized_headers
 
     # @return [String] the request body, up to a maximum of 16KiB. If the
     #   underlying IO object had been partially (or fully) read before,
-    #   this will restore the previous cursor position after reading it.
+    #   this will restore the previous cursor position after reading it if
+    #   possible, or left it rewund if not.
     def truncated_body(max_size: 16384)
```
willem-delbare, line 44: what doe we use the truncated body for?

foca: This is only used for the attack events, where we want to include the request data (including the body) in the JSON payload.

foca: Improved the documentation for this.

PR #21, opened 2024-08-29
Track outbound connections

foca:
This sets up the code to track outbound connections in any sink that wraps a library that does such a thing, setting it up with `net-http` (the stdlib's HTTP library) for now.

This also refactors the `Stats#serialize_and_reset` which was getting unwieldy, by just creating a copy of the entire stats object before resetting it, and returning the copy. This simplifies passing it down to the heartbeat event, rather than having to serialize a bunch of things and passing the serialized stuff one by one.

### DIFF
--- a/lib/aikido/firewall/sink.rb
+++ b/lib/aikido/firewall/sink.rb
@@ -51,7 +51,9 @@ module Aikido::Firewall
     #   using an +operation+ param of their own.
     attr_reader :operation
 
-    def initialize(name, scanners:, operation: name, reporter: Aikido::Agent.method(:track_scan))
+    DEFAULT_REPORTER = ->(scan) { Aikido::Agent.track_scan(scan) }
+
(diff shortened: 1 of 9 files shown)


lib/aikido/agent/stats.rb
```diff
@@ -126,6 +149,78 @@ def routes_as_json
...
+    # @api private
+    #
+    # Keeps track of the visited routes.
+    class Routes
+      def initialize
+        @routes = Hash.new(0)
+      end
+
+      # @param route [Aikido::Agent::Route, nil] tracks the visit, if given.
+      def add(route)
```
willem-delbare: this class is becoming a big boy

foca: The file is long, but it's really 4 classes and they all have a small-ish interface.

I'm not against breaking it up and moving things to separate files, but for things that are only used once in a single place, I tend to leave things together because it makes it easier to follow the code when you are just scanning up and down rather than having to jump between several files. At least for me!

PR #32, opened 2024-09-05
API for user tracking

foca:
Developers can now call `Aikido::Agent.track_user(id: <id>, name: <name>)` from their code whenever a User either signs in or is set as the request's current user (i.e. by reading a cookie, or an Authorization header).

In order to keep the API simple for devs, we'll do a best guess to map the object given to an "actor":

* ~If the user is an ActiveModel-like object (i.e. implements `#to_model`), we'll read the `id` and `name` attributes (configurable) from the model directly.~ (we're not doing dynamic dispatching of attributes anymore)

* If they pass a Hash, we'll expect either `:id`/`:name` or `"id"`/`"name"` to be keys, and use those values.

* If they pass an object that implements `#to_aikido_actor`, we'll call that and expect that the method returns an `Aikido::Agent::Actor`. The idea is to give users a simple protocol they can implement when they have complex requirements.

The last case is meant for situations where simply mapping the ID can prove insufficient, like when multiple different types of models are used to authenticate—for example, a `User` and `Admin`, which could have clashing IDs (i.e. both have autoincrementing numeric IDs).

This lets users do something like:

``` ruby
class Admin < ApplicationRecord
  def to_aikido_actor
    Aikido::Agent::Actor.new(id: "admin:#{id}", name: email)
  end
end
```

So they see in the UI the ID `1` and know it's the user, and the ID `admin:1` and know it's the admin.

### DIFF
--- a/lib/aikido/agent/actor.rb
+++ b/lib/aikido/agent/actor.rb
@@ -11,13 +11,6 @@ module Aikido::Agent
   #     will have that method called and its value returned.
   #   @return Aikido::Agent::Actor
   #
-  # @overload Actor(model)
-  #   @param model [#to_model] any ActiveModel object that implements #to_model
-  #     will have that method called, and the result will have its attributes
-  #     mapped via {Aikido::Agent::Config#user_attribute_mappings} to build an
-  #     Actor object.
-  #   @return Aikido::Agent::Actor
-  #
   # @overload Actor(data)
   #   @param data [Hash<Symbol, String>]
   #   @option data [String] :id a unique identifier for this user.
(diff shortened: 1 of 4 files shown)


lib/aikido/agent/actor.rb
```diff
@@ -0,0 +1,132 @@
...
+    return if data.nil?
+    return data.to_aikido_actor if data.respond_to?(:to_aikido_actor)
+
+    attrs = {}
+    if data.respond_to?(:to_model)
+      model = data.to_model
+      id_attr, name_attr = config.user_attribute_mappings.values_at(:id, :name)
+
+      # Normalize objects with composite keys whose #id might be an Array.
+      id = Array(model.public_send(id_attr))&.join(":")
```
willem-delbare: might get flagged

PR #41, opened 2024-10-22
Release scripts

foca:
Sets up a release pipeline that will automatically push to rubygems whenever we create a Release on GitHub.

This will build "native gems" for each architecture we support, bundling the specific libzen_internals to that architecture, to keep size down (compressed `.gem` files are ~1mb each).

Then, when you install the gem, it will select the appropriate architecture and download that one automatically:

Screenshot 2024-10-24 at 14 40 17

(the `--pre` is because we've only pushed a prerelease version of the gem so far, and so we need that flag for it to be installed.)

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -3,7 +3,6 @@ name: Release
 on:
   release:
     types: [published]
-  workflow_dispatch:
 
 jobs:
   build:


.github/workflows/release.yml
```diff
@@ -0,0 +1,28 @@
...
+      id-token: write
+
+    needs:
+      - build
+    steps:
+      - uses: actions/checkout@v4
+      - name: Set up Ruby
+        uses: ruby/setup-ruby@v1
+        with:
+          ruby-version: ruby
```
hansott, line 27: is this correct? should be a version? like in the main workflow?

foca: It's weird but correct :)

The way this interprets ruby versions it has two parts to it: The engine (CRuby, JRuby, TruffleRuby, mruby, …) and the version number. If you wanted to fully qualify a version, you'd actually say `ruby-3.3.4` to specify CRuby 3.3.4. Omitting the version just means "the latest available version".

Because this is only for building the gem, we can just say "use the latest", we don't really care about having a matrix of versions or anything, and it's one less place to remember to update as new versions come out :) #lazy

.github/workflows/release.yml
```diff
@@ -0,0 +1,30 @@
+name: Release
+
+on:
+  release:
+    types: [published]
+  workflow_dispatch:
```
willem-delbare: is this correct?

foca: Yep. YAML syntax is weird, but it just means it's an empty hash. Which it has to be because `release` is a hash.

PR #69, opened 2025-02-19
Add 3 more methods to detect attacks

HiDAl:
3 more methods (`File.join, File.chown, File.chmod`) are monkey patched, so we can detect potential attacks.

Also, refactored the tests to make more clear and simpler adding methods.

### DIFF
--- a/benchmarks/rails7.1_sql_injection.js
+++ b/benchmarks/rails7.1_sql_injection.js
@@ -1,8 +1,5 @@
 import http from 'k6/http';
-import { textSummary } from 'https://jslib.k6.io/k6-summary/0.0.2/index.js';
-import { check, sleep, fail } from 'k6';
-import exec from 'k6/execution';
-import { Trend } from 'k6/metrics';
+import {Trend} from 'k6/metrics';
 
 const HTTP = {
   withZen: {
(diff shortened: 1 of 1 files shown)


lib/aikido/zen/sinks/file.rb
```diff
@@ -21,13 +21,27 @@ def self.scan_path(filepath, operation)
...
           super
         end
 
         def write(filename, *, **)
           Extensions.scan_path(filename, "write")
+          super
+        end
+
+        def join(*)
+          joined = super
```
bitterpanda63, line 33: We need to make sure that this joined path still includes the "../" right? That was an issue in e.g. nodejs that the paths were already resolved then

HiDAl: [here](https://github.com/AikidoSec/firewall-ruby/pull/69/files#diff-47f65382d057ebbce7b6195d48940e81097b5d9c4a06b1927bc4a7ad51577adbR69) is a test checking that the joined path still contains `../`... not fully sure that I understood the concern.

bitterpanda63: Awesome

PR #71, opened 2025-02-21
Fix benchmarks + refactor

HiDAl:
while checking the benchmark, I noticed we weren't disabling Zen due to a typo. This PR fix that typo and also refactor a bit the code, improving naming of the measured metrics and exposing more metrics

Screenshot 2025-02-21 at 14 08 29

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -137,7 +137,5 @@ jobs:
       run: "bundle exec rake libzen:download:current"
     - name: Run the e2e tests
       run: "bundle exec rake e2e:${{ matrix.app }}"
-    - name: Prepare benchmarks
-      run: "bundle exec rails assets:precompile"
     - name: Run the benchmarks
       run: "bundle exec rake bench:${{ matrix.app }}:run"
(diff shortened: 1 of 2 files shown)


benchmarks/rails7.1_sql_injection.js
```diff
@@ -13,51 +13,44 @@ const HTTP = {
...
-    test_get_page_without_attack: ["med<10"],
-    test_get_page_with_sql_injection: ["med<10"],
+    test_post_page_with_json_body_delta: ["med<10"],
+    test_get_page_without_attack_delta: ["med<10"],
+    test_get_page_with_sql_injection_delta: ["med<10"],
   }
 };
 
-const expectAttack = http.expectedStatuses(500);
+const expectAttack = http.expectedStatuses(200, 500);
```
timokoessler, line 53: Why do you expect status code 200 in case of an attack?

HiDAl: in the case Zen is disabled, the response is `200` because the attack wasn't stopped.

timokoessler: Ah, okay. Would maybe be safer to separate this, so that it would fail if 200 is returned and zen is enabled?

HiDAl: yeah, that refactor makes sense, but implies change several stuff of this benchmark. I'll work on it later.

PR #73, opened 2025-02-26
Add support to detect shell injection attacks

HiDAl:
Ported over all the paraphernalia from `AikidoSec/firewall-node` Shell Injection to Ruby.


lib/aikido/zen/attack.rb
```diff
@@ -72,6 +72,39 @@ def exception(*)
...
+
+      def initialize(input:, command:, **opts)
+        super(**opts)
+        @input = input
+        @command = command
+      end
+
+      def log_message
+        format(
+          "Shell Injection: Malicious user input «%s» detected while calling method %s",
```
bitterpanda63, line 87: Let's rm user input log, could be used for log injection, and it's unnecessary

bitterpanda63: this is gonna happen in separate PR

PR #74, opened 2025-02-26
Refactor attack logging logic

HiDAl:
push all the way up to the Agent the duty to create the `attack detected` message, same than the message logged by agents in other languages.

### DIFF
--- a/lib/aikido/zen/agent.rb
+++ b/lib/aikido/zen/agent.rb
@@ -104,13 +104,7 @@ module Aikido::Zen
       attack.will_be_blocked! if @config.blocking_mode?
 
       @config.logger.error(
-        format('Zen has %s a %s: kind="%s" operation="%s(...)" source="%s%s"',
-          attack.blocked? ? "blocked" : "detected",
-          attack.humanized_name,
-          attack.kind,
-          attack.operation,
-          attack.input.source,
-          attack.input.path).delete("\n")
+        format("Zen has %s a %s: %s", attack.blocked? ? "blocked" : "detected", attack.humanized_name, attack.as_json.to_json)
       )
       report(Events::Attack.new(attack: attack)) if @api_client.can_make_requests?
 
--- a/test/aikido/zen/agent_test.rb
+++ b/test/aikido/zen/agent_test.rb
@@ -206,7 +206,7 @@ class Aikido::Zen::AgentTest < ActiveSupport::TestCase
     @agent.handle_attack(attack)
 
     assert_logged :error,
-      /Zen has detected a humanized test attack name: kind="test_attack" operation="test\(...\)" source="sourcepath/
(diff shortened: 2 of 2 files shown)


lib/aikido/zen/agent.rb
```diff
@@ -103,7 +103,15 @@ def updated_settings!
...
 
-      @config.logger.error("[ATTACK DETECTED] #{attack.log_message}")
+      @config.logger.error(
+        format('Zen has %s a %s: kind="%s" operation="%s(...)" source="%s%s"',
+          attack.blocked? ? "blocked" : "detected",
+          attack.humanized_name,
+          attack.kind,
+          attack.operation,
+          attack.input.source,
+          attack.input.path).delete("\n")
```
bitterpanda63: what about tabs, \r, etc...

lib/aikido/zen/attack.rb
```diff
@@ -49,22 +66,18 @@ def initialize(input:, filepath:, **opts)
...
       end
 
-      def log_message
-        format(
-          "Path Traversal: Malicious user input «%s» detected while calling method %s",
-          @input, @operation
-        )
+      def metadata
+        {
+          expanded_filepath: filepath
```
bitterpanda63: To store on aikido's servers the metadata needs to line up with the other agents, I think the other agens just use "filepath" as property name, but I would double check

lib/aikido/zen/agent.rb
```diff
@@ -103,7 +103,15 @@ def updated_settings!
     def handle_attack(attack)
       attack.will_be_blocked! if @config.blocking_mode?
 
-      @config.logger.error("[ATTACK DETECTED] #{attack.log_message}")
+      @config.logger.error(
```
bitterpanda63, line 106: would not format, just output stringified json? Because thats the easiest for people who e.g. want to run Zen without Aikido or to analyze the attack (like the payload and metadata as well)

HiDAl: probably we should kick off the conversations about unifying logging. The nodeJS agent is using plain strings too https://github.com/AikidoSec/firewall-node/blob/5db7d7fdea31359be9e6bbd992ce97bb82ea559d/library/agent/AttackLogger.ts#L33

lib/aikido/zen/attack.rb
HiDAl: right. will fix it

HiDAl: `filename` actually https://github.com/AikidoSec/firewall-node/blob/5db7d7fdea31359be9e6bbd992ce97bb82ea559d/library/vulnerabilities/path-traversal/checkContextForPathTraversal.ts#L43

lib/aikido/zen/agent.rb
HiDAl: weird... was sure about using tabs and such. will fix it

PR #75, opened 2025-02-27
Add support to detect shell injection attacks

HiDAl:
**IMPORTANT:** Patching the `Kernel` module in ruby < 3.0 is not possible, so we are not bringing support to that version

### DIFF
--- a/test/aikido/zen/sinks/kernel_test.rb
+++ b/test/aikido/zen/sinks/kernel_test.rb
@@ -21,8 +21,6 @@ class Aikido::Zen::Sinks::KernelTest < ActiveSupport::TestCase
   end
 
   test "system works normally" do
-    skip_if_ruby_lower_than "3.0"
-
     assert system("echo")
     assert system(SOME_ENV, "echo")
     assert system("echo", unsetenv_others: true)
@@ -44,8 +42,6 @@ class Aikido::Zen::Sinks::KernelTest < ActiveSupport::TestCase
   end
 
   test "spawn works normally" do
-    skip_if_ruby_lower_than "3.0"
-
     def assert_spawn_runs_normally(&block)
       assert_nothing_raised do
         pid = yield


lib/aikido/zen/scanners/shell_injection_scanner.rb
```diff
@@ -2,59 +2,61 @@
 
 require_relative "shell_injection/helpers"
 
-module Aikido::Zen::Scanners
```
bitterpanda63, line 5: Oof why so much refactoring, what's new here?

lib/aikido/zen/sinks.rb
```diff
@@ -6,6 +6,13 @@
 
 require_relative "sinks/action_controller" if defined?(::ActionController)
 require_relative "sinks/file" if defined?(::File)
+
+# Sadly, in ruby versions lower than 3.0, it's not possible to patch the
+# Kernel module because how the `prepend` method is applied
+# (https://stackoverflow.com/questions/78110397/prepend-kernel-module-function-globally#comment137713906_78112924)
+if RUBY_ENGINE >= "3.0"
```
bitterpanda63: What's the difference between `RUBY_ENGINE` and `RUBY_VERSION`

lib/aikido/zen/sinks/kernel.rb
```diff
@@ -0,0 +1,74 @@
...
+        # Checks if the user introduced input is trying to execute other commands
+        # using Shell Injection kind of attacks.
+        #
+        # @param command [String] the _full command_ that will be executed.
+        # @param context [Aikido::Zen::Context]
+        # @param sink [Aikido::Zen::Sink] the Sink that is running the scan.
+        # @param operation [Symbol, String] name of the method being scanned.
+        #
+        # @return [Aikido::Zen::Attacks::ShellInjectionAttack, nil] an Attack if any
+        # user input is detected to be attempting a Shell Injection Attack, or +nil+ if not.
```
bitterpanda63:
```suggestion
        # user input is detected as part of a Shell Injection, or +nil+ if it's safe.
```

lib/aikido/zen/sinks/kernel.rb
```diff
@@ -0,0 +1,74 @@
...
+        def self.scan_command(command, operation)
+          SINK.scan(
+            command: command,
+            operation: operation
+          )
+        end
+
+        # `system, spawn` functions can be invoked in several ways. For more details,
+        # see [the documentation](https://apidock.com/ruby/Kernel/spawn)
```
bitterpanda63: Would we not want to use official docs, or is the apidock.com more common?
```suggestion
        # see [the documentation](https://ruby-doc.org/3.4.1/Kernel.html#method-i-spawn)
```

test/test_helper.rb
```diff
@@ -124,6 +125,12 @@ def assert_hash_subset_of(container, subset)
     assert_equal container.slice(*subset.keys), subset
   end
 
+  def skip_if_ruby_lower_than(expected_ruby_version)
```
bitterpanda63, line 128: Do we also test for lower versions that no scans happen, just to verify the behaviour?

lib/aikido/zen/sinks/kernel.rb
```diff
@@ -0,0 +1,74 @@
...
+        #
+        # In our context, we care primarily about two common scenarios:
+        #   - one argument (String)
+        #       e.g.: system("ls"), system("echo something")
+        #   - two arguments (Hash, String)
+        #       e.g.: system({"foo" => "bar"}, "ls"), system({"foo" => "bar"}, "echo something")
+        #
+        # In all other cases, Ruby's default behavior ensures that user input is appropriately
```
bitterpanda63: would simply mention we don't protect against shell argument injections right, we don't want to make promises in code that it's fully safe

lib/aikido/zen/scanners/shell_injection_scanner.rb
HiDAl: only whitespaces to follow the same structure that other scanners defined like this:

```ruby


module Aikido::Zen
  module Scanners
    module <ModuleName>
```

lib/aikido/zen/sinks.rb
HiDAl: my bad, should be `RUBY_VERSION`

test/test_helper.rb
HiDAl: you're right. Fixed it uncommenting the first 2 tests.

PR #76, opened 2025-02-28
AIK-4137: Port tracking logic from node

HiDAl:
Ports over all the logic to track URLs from [firewall-node](https://github.com/AikidoSec/firewall-node/blob/main/library/sources/http-server/shouldDiscoverRoute.ts)

### DIFF
--- a/lib/aikido/zen/middleware/request_tracker.rb
+++ b/lib/aikido/zen/middleware/request_tracker.rb
@@ -32,6 +32,8 @@ module Aikido::Zen
       # @param route [String]
       # @param http_method [String]
       def track?(status_code:, route:, http_method:)
+        # In the UI we want to show only successful (2xx) or redirect (3xx) responses
+        # anything else is discarded.
         return false unless status_code >= 200 && status_code <= 399
 
         return false if IGNORED_METHODS.include?(http_method)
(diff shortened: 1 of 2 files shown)


lib/aikido/zen/rails_engine.rb
```diff
@@ -14,6 +14,7 @@ class RailsEngine < ::Rails::Engine
 
       app.middleware.use Aikido::Zen::Middleware::SetContext
       app.middleware.use Aikido::Zen::Middleware::CheckAllowedAddresses
+      app.middleware.use Aikido::Zen::Middleware::RequestTracker
```
bitterpanda63, line 19: does this mean its the first aikido middleware to run or the last, because the request tracker should be at the absolute end of the request (i.e. final status code) would add comment explaining that

lib/aikido/zen.rb
```diff
@@ -70,6 +72,11 @@ def self.track_request(request)
       collector.track_request(request)
     end
 
+    def self.track_discovered_route(request)
```
bitterpanda63, line 75: okay and this only gets called from that middleware after we've checked that we want to discover the route

lib/aikido/zen/middleware/request_tracker.rb
```diff
@@ -0,0 +1,90 @@
...
+
+      IGNORED_METHODS = %w[OPTIONS HEAD]
+      IGNORED_EXTENSIONS = %w[properties config webmanifest]
+      IGNORED_SEGMENTS = ["cgi-bin"]
+
+      # @param status_code [Integer]
+      # @param route [String]
+      # @param http_method [String]
+      def track?(status_code:, route:, http_method:)
+        return false unless status_code >= 200 && status_code <= 399
```
bitterpanda63, line 37: lets add comment explaining we dont want 404/failed routes showing up in dashboard

PR #82, opened 2025-03-06
Implement blocking user logic according user's configuration

HiDAl:
Implements all the logic to honor blocked users in Aikido's dashboard.

### DIFF
--- a/lib/aikido/zen/sinks/action_controller.rb
+++ b/lib/aikido/zen/sinks/action_controller.rb
@@ -3,12 +3,12 @@
 module Aikido::Zen
   module Sinks
     module ActionController
-      # Implements the "middleware" for guard the gate (i.e.: rate-limiting or blocking)
-      # in Rails apps, where we need to check at the end of the `before_action` chain,
-      # rather than in an actual Rack middleware, to allow for calls to Zen.track_user being
-      # made from before_actions in the host app, thus allowing guard the gate  by user ID
-      # rather than solely by IP.
-      class GateGuard
+      # Implements the "middleware" for blocking requests (i.e.: rate-limiting or blocking
+      # user/bots) in Rails apps, where we need to check at the end of the `before_action`
+      # chain, rather than in an actual Rack middleware, to allow for calls to
+      # `Zen.track_user` being made from before_actions in the host app, thus allowing
+      # block/rate-limit by user ID rather than solely by IP.
+      class BlockRequestChecker
         def initialize(
           config: Aikido::Zen.config,
           settings: Aikido::Zen.runtime_settings,
@@ -19,7 +19,7 @@ module Aikido::Zen
           @rate_limiter = rate_limiter
         end
 
-        def block_pass?(controller)
+        def block?(controller)
           context = controller.request.env[Aikido::Zen::ENV_KEY]
           request = context.request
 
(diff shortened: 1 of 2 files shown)


lib/aikido/zen/config.rb
```diff
@@ -236,9 +236,14 @@ def read_boolean_from_env(value)
...
 
     # @!visibility private
-    DEFAULT_BLOCKED_IP_RESPONDER = ->(request) do
-      message = "Your IP address is not allowed to access this resource. (Your IP: %s)"
-      [403, {"Content-Type" => "text/plain"}, [format(message, request.ip)]]
+    DEFAULT_BLOCKED_RESPONDER = ->(request, blocking_type) do
+      message = case blocking_type
+      when :ip
+        format("Your IP address is not allowed to access this resource. (Your IP: %s)", request.ip)
+      else
```
bitterpanda63, line 243: With bot blocking we show a different message, is bot blocking also using this? See node response for bot blockign

lib/aikido/zen/sinks/action_controller.rb
```diff
@@ -8,7 +8,7 @@ module ActionController
       # an actual Rack middleware, to allow for calls to Zen.track_user being
       # made from before_actions in the host app, thus allowing rate-limiting
       # by user ID rather than solely by IP.
-      class Throttler
+      class GateGuard
```
bitterpanda63: What's GateGuard, non-obv now

lib/aikido/zen/sinks/action_controller.rb
```diff
@@ -34,26 +42,34 @@ def throttle(controller)
           false
         end
 
-        private def should_throttle?(request)
+        private
+
+        def should_throttle?(request)
```
bitterpanda63: revert

lib/aikido/zen/sinks/action_controller.rb
bitterpanda63: Comment still calls it Middleware, why not ZenDefaultMiddleware or something along those lines

lib/aikido/zen/config.rb
HiDAl: this agents has no support for bot blocking. When we implement that, we have to add that message

lib/aikido/zen/sinks/action_controller.rb
HiDAl: ??

lib/aikido/zen/sinks/action_controller.rb
HiDAl: this is not the default middleware. I couldn't come with a better name though. Let me think a bit more

lib/aikido/zen/config.rb
bitterpanda63: right sorry.

lib/aikido/zen/sinks/action_controller.rb
bitterpanda63: why did we add whitespace?

HiDAl: ok ok , will revert

HiDAl: sorry, in ruby you can mark as private a method or all the methods after calling the `private` function. 
Hence, it was my intention to have this `private` here, because it's marking the following 2 methods —`should_throttle? & should_block_user?`— as private.

PR #83, opened 2025-03-06
Add `middleware_installed` information to heartbeats

HiDAl:
Sample ruby does not have the banner indicating the middleware must be installed:

Screenshot 2025-03-06 at 17 19 52

### DIFF
--- a/lib/aikido/zen/sinks/action_controller.rb
+++ b/lib/aikido/zen/sinks/action_controller.rb
@@ -3,12 +3,12 @@
 module Aikido::Zen
   module Sinks
     module ActionController
-      # Implements the "middleware" for rate limiting in Rails apps, where we
-      # need to check at the end of the `before_action` chain, rather than in
-      # an actual Rack middleware, to allow for calls to Zen.track_user being
-      # made from before_actions in the host app, thus allowing rate-limiting
-      # by user ID rather than solely by IP.
-      class GateGuard
+      # Implements the "middleware" for blocking requests (i.e.: rate-limiting or blocking
+      # user/bots) in Rails apps, where we need to check at the end of the `before_action`
+      # chain, rather than in an actual Rack middleware, to allow for calls to
+      # `Zen.track_user` being made from before_actions in the host app, thus allowing
+      # block/rate-limit by user ID rather than solely by IP.
+      class BlockRequestChecker
         def initialize(
           config: Aikido::Zen.config,
           settings: Aikido::Zen.runtime_settings,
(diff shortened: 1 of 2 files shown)


lib/aikido/zen/middleware/rack_throttler.rb
```diff
@@ -23,6 +23,8 @@ def initialize(
       def call(env)
         request = Aikido::Zen::Middleware.request_from(env)
 
+        Aikido::Zen.middleware_installed!
```
timokoessler, line 26: Middleware installed is set to true at two different positions?

lib/aikido/zen.rb
```diff
@@ -120,6 +120,12 @@ def self.track_user(user)
       end
     end
 
+    # Marks that the Zen middleware was installed properly
```
bitterpanda63, line 123: who uses this code and why?

lib/aikido/zen/middleware/rack_throttler.rb
bitterpanda63: this should be the Collector class right

lib/aikido/zen/sinks/action_controller.rb
```diff
@@ -20,6 +20,8 @@ def initialize(
         end
 
         def block_pass?(controller)
+          Aikido::Zen.middleware_installed!
```
bitterpanda63, line 23: Same here why not the Collector class

lib/aikido/zen/middleware/rack_throttler.rb
HiDAl: we have 2 middlewares:
- first one for normal rack
- second specific for rails

why? because in the lifecycle of rails request, we can't easily access to the current user stored in the context. That's why we have to patch rails lifecycle with this `ActionController` stuff. This allows our users to call the method `track_user(...)` within the `before_action: ` function, making available in the Zen-request-context the user information.

HiDAl: In the end, we'll have to write custom logic, framework-by-framework, patching the lifecycle to properly have the current user id.

lib/aikido/zen/sinks/action_controller.rb
HiDAl: because collector is internal and private to `Aikido::Zen::Agent`

lib/aikido/zen.rb
HiDAl: this is the public API of the agent

PR #84, opened 2025-03-07
Marks that middleware is installed by the end of rail's bootstraping process

HiDAl:


lib/aikido/zen/rails_engine.rb
```diff
@@ -60,6 +60,8 @@ class RailsEngine < ::Rails::Engine
       # that any gems required after aikido-zen are detected and patched
       # accordingly.
       Aikido::Zen.load_sinks!
+
+      Aikido::Zen.middleware_installed!
```
bitterpanda63, line 67: add comment explaining why though, and that this is Rails specific?

PR #86, opened 2025-03-18
refactoring: Break the circular dependencies between internal services

HiDAl:
This PR is the first one of a serie of PR doing a major refactor to support multi-worker environments.

* Force to explicit start the agent in the framework loading (users won't have to to anything)
* fix a bug where we were starting an agent only to stop it (affecting tests)
* Some object were sending messages to their parent, creating circular dependencies and making the agent to be started in tests (i.e.: RuntimeSettings)
* Agent is not started anymore to runt tests

### DIFF
--- a/test/test_helper.rb
+++ b/test/test_helper.rb
@@ -52,6 +52,17 @@ class ActiveSupport::TestCase
     Aikido::Zen.instance_variable_set(:@runtime_settings, nil)
     Aikido::Zen.current_context = nil
 
+    Aikido::Zen.singleton_class.remove_method(:track_scan)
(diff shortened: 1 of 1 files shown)


lib/aikido/zen/runtime_settings.rb
```diff
@@ -56,7 +56,7 @@ def update_from_json(data)
       self.skip_protection_for_ips = RuntimeSettings::IPSet.from_json(data["allowedIPAddresses"])
       self.received_any_stats = data["receivedAnyStats"]
 
-      Aikido::Zen.agent.updated_settings! if updated_at != last_updated_at
```
bitterpanda63, line 59: why was this removed, seems important to check if smth was updated or am I wrong?

PR #87, opened 2025-03-19
Introduce Ruby remote calls DRB

HiDAl:
Ruby provides a built-in library for remote procedure calls [(RDb)](https://en.wikibooks.org/wiki/Ruby_Programming/Standard_Library/DRb), which is simpler and more lightweight compared to using gRPC. It's entirely implemented in Ruby, meaning there are no external dependencies required.

So far, implementing it took just an hour, and it’s definitely easier to handle than the complexity of gRPC.

### DIFF
--- a/lib/aikido/zen/detached_agent.rb
+++ b/lib/aikido/zen/detached_agent.rb
@@ -1,6 +1,7 @@
 # frozen_string_literal: true
 
 require "drb/drb"
+require "drb/unix"
 
 module Aikido::Zen
   # It's possible to use `extend Forwardable` here for one-line forward calls to the


lib/aikido/zen/sinks/socket.rb
```diff
@@ -15,6 +15,12 @@ module Socket
 
       module IPSocketExtensions
         def self.scan_socket(hostname, socket)
+          # This scanner is checking remote connections. It tries to
+          # gather data from the remote connection using peeraddr.
+          # In case it's a TCPServer, it's not connecting to any remote
+          # address, so the scan is not needed.
+          return if socket.is_a?(TCPServer)
```
bitterpanda63: Maybe also UDPServer, other types of servers?

lib/aikido/zen/config.rb
```diff
@@ -153,6 +156,7 @@ def initialize
       self.debugging = read_boolean_from_env(ENV.fetch("AIKIDO_DEBUG", false))
       self.logger = Logger.new($stdout, progname: "aikido", level: debugging ? Logger::DEBUG : Logger::INFO)
       self.max_performance_samples = 5000
+      self.detached_agent_socket_path = "aikido-detached-agent.socket"
```
bitterpanda63, line 161: where is this file then created? In python/php we use a separate folder, one that users can also change if they like. But maybe Ruby handles the creation of this file in it's own separate folder? Would at least add a comment explaining that

lib/aikido/zen/config.rb
```diff
@@ -210,6 +214,11 @@ def api_timeouts=(value)
       @api_timeouts.update(value)
     end
 
+    def detached_agent_socket_path=(path)
+      @detached_agent_socket_path = path
+      @detached_agent_socket_path = "drbunix:" + @detached_agent_socket_path unless @detached_agent_socket_path.start_with?("unix:")
```
bitterpanda63: what's the difference between drbunix: and unix:, a bit weird let's add comments or fix

lib/aikido/zen.rb
```diff
@@ -144,8 +153,26 @@ def self.agent
       @agent ||= Agent.start
     end
 
+    def self.detached_agent_server
+      @detached_agent_server ||= DetachedAgentServer.start!
+    end
+
     class << self
-      alias_method :start!, :agent
+      LOCK = Mutex.new
```
bitterpanda63, line 163: let's add a comment explaining why we create a mutex here

lib/aikido/zen/detached_agent.rb
```diff
@@ -0,0 +1,77 @@
...
+      @detached_agent_front.track_request
+    end
+
+    def middleware_installed!
+      @detached_agent_front.middleware_installed!
+    end
+
+    # Everytime a fork happens, we have to start a DRb service in a background thread.
+    # It will be in charge of maintaining the connection and clean up resources.
+    def handle_fork
```
bitterpanda63, line 28: can we explicitly clarify fork == new process and != new thread?

lib/aikido/zen/detached_agent.rb
```diff
@@ -0,0 +1,77 @@
...
+      DRb.start_service
+    end
+  end
+
+  class DetachedAgentFront
```
bitterpanda63, line 34: What does `Front` mean exactly, let's rename or specify

lib/aikido/zen/detached_agent.rb
```diff
@@ -0,0 +1,77 @@
...
+      @config = config
+      @collector = collector
+    end
+  end
+
+  class DetachedAgentServer
+    def initialize(config: Aikido::Zen.config)
+      @detached_agent_front = DetachedAgentFront.new
+      @drb_server = DRb.start_service(config.detached_agent_socket_path, @detached_agent_front)
+      @drb_server.verbose = config.logger.debug?
```
bitterpanda63, line 51:
```suggestion
      @drb_server.verbose = config.logger.debug? # We don't want to see drb logs unless in debug mode
```

lib/aikido/zen/detached_agent.rb
```diff
@@ -0,0 +1,77 @@
...
+
+    def stop!
+      @drb_server.stop_service
+      DRb.stop_service
+    end
+
+    class << self
+      def start!
+        Aikido::Zen.config.logger.debug("Starting DRb Server...")
+        max_attempts = 10
```
bitterpanda63, line 66: why exactly are we working with 10 attempts, what's happening so that the drb server cannot start one time but can start the other time?

lib/aikido/zen/config.rb
HiDAl: internal Ruby stuff. drb = protocol / unix = file socket.

lib/aikido/zen/detached_agent.rb
HiDAl: Front from "front object". A dRB convention https://docs.ruby-lang.org/en/3.0/DRb.html

lib/aikido/zen/detached_agent.rb
HiDAl: why? isn't explicit the idea here?

lib/aikido/zen/config.rb
HiDAl: users can change this configuration. My idea is to updated it to the frameworks tmp folder (rails has this `tmp` folder, where they put sockets, caches, etc). 
Right now, it's in the root of the application.

lib/aikido/zen/detached_agent.rb
HiDAl: Starting the server is a fire-and-forget action (it runs in a separate thread). Since it might take longer to start than expected, we give it some extra time.

lib/aikido/zen/config.rb
bitterpanda63: yeah best to use tmp folder by default and allow config change if necessary 👍🏻 but ok next PR

PR #88, opened 2025-03-19
Add support to collect metrics in the dRB process

HiDAl:


lib/aikido/zen/request/rails_router.rb
```diff
@@ -58,27 +58,15 @@ def recognize(request)
...
-  module Rails
-    class Route < Aikido::Zen::Route
-      attr_reader :verb
+      route_wrapper = ActionDispatch::Routing::RouteWrapper.new(route)
 
-      def initialize(rails_route, verb: rails_route.verb, prefix: nil)
-        @route = ActionDispatch::Routing::RouteWrapper.new(rails_route)
-        @verb = verb
-        @prefix = prefix
+      path = if prefix.present?
```
HiDAl, line 63: Simplifies the creation of `Route` objects, otherwise, we'll need to have different Structs in `Detached` agent. Also, this wasn't needed at all, just a wrapper for the route object

lib/aikido/zen/request/rails_router.rb
```diff
@@ -58,27 +58,15 @@ def recognize(request)
...
+        File.join(prefix.to_s, route_wrapper.path).chomp("/")
```
timokoessler, line 64: Is it safe to use file related apis for http url routes? E.g. on Windows with `\`?

lib/aikido/zen.rb
bitterpanda63, line 1: Can we remove import for zen/collector, what is that file doing now anyways?

lib/aikido/zen/agent.rb
```diff
@@ -28,6 +29,7 @@ def initialize(
       @worker = worker
       @api_client = api_client
       @collector = collector
```
bitterpanda63, line 31: Still doing smth useful? the name is a bit weird now, not sur what it does

lib/aikido/zen/request/schema/auth_schemas.rb
```diff
@@ -18,6 +18,20 @@ def as_json
         @schemas.map(&:as_json) unless @schemas.empty?
       end
 
+      def self.from_json(schemas_array)
```
bitterpanda63, line 21: why is this related, is it bc we need to send primitive types over rdb?

HiDAl: yes, we need to transform it in some way

lib/aikido/zen/agent.rb
HiDAl: it's still used by the detached agent. we just moved the responsibility to store the metrics to the other agent

lib/aikido/zen.rb
HiDAl: no, it's still used

lib/aikido/zen/detached_agent.rb
```diff
@@ -34,12 +68,46 @@ def handle_fork
...
+
     def_delegators :@collector, :middleware_installed!, :track_request
 
     def initialize(config: Aikido::Zen.config, collector: Aikido::Zen.collector)
       @config = config
       @collector = collector
     end
+
+    def track_route(route, schema)
+      @collector.track_route(RequestKind.new(route, Aikido::Zen::Request::Schema.from_json(schema)))
```
bitterpanda63, line 93: we have 3x funcs with the same name now for all these tracking funcs, but that can take different data, any way we can make this clearer/simplify?
Like e.g. is the @collector now only used in the agent thread?

lib/aikido/zen/request/rails_router.rb
HiDAl: to be honest, no idea. This is old code, I just refactored it a bit. Do you have an idea about why would it be unsafe?

timokoessler: On Windows it might use \ instead of /? That would happen in Node. It seems like Ruby always uses /? https://github.com/ruby/ruby/blob/fc26004660a287b977e5481a09f6329ffa688357/file.c#L5120

bitterpanda63: Created an issue in Jira for this https://aikido.atlassian.net/browse/AIK-4881

PR #89, opened 2025-03-20
Add support to avoid running scanners when context is `nil`

HiDAl:


lib/aikido/zen/scanners/ssrf_scanner.rb
```diff
@@ -6,6 +6,10 @@
 module Aikido::Zen
   module Scanners
     class SSRFScanner
+      def self.skips_on_nil_context?
+        true
```
bitterpanda63, line 12: explain why exactly here though

lib/aikido/zen/sink.rb
```diff
@@ -84,11 +84,16 @@ def scan(context: Aikido::Zen.current_context, **scan_params)
 
       scan = Scan.new(sink: self, context: context)
 
+      scans_performed = 0
```
bitterpanda63, line 87: why is it related to the nil check

bitterpanda63: nevermind

PR #90, opened 2025-03-20
load agent after the sinks are loaded

HiDAl:
Otherwise we'll no have the packages when reporting to aikido's backend

### DIFF
--- a/lib/aikido/zen/rails_engine.rb
+++ b/lib/aikido/zen/rails_engine.rb
@@ -63,6 +63,7 @@ module Aikido::Zen
 
       # It's important we start after loading sinks, so we can report the installed packages
       Aikido::Zen.start!
+      Aikido::Zen.start!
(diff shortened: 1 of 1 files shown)


lib/aikido/zen/rails_engine.rb
```diff
@@ -56,13 +56,13 @@ class RailsEngine < ::Rails::Engine
...
       end
 
-      Aikido::Zen.start!
-
       # Make sure this is run at the end of the initialization process, so
       # that any gems required after aikido-zen are detected and patched
       # accordingly.
       Aikido::Zen.load_sinks!
 
+      Aikido::Zen.start!
```
bitterpanda63, line 66:
```suggestion
      # It's important we start after loading sinks, so we can report the installed packages
      Aikido::Zen.start!
```

PR #94, opened 2025-03-27
Process monitoring in a Background Worker

HiDAl:
Detached agent buffers all the stats per worker and each 10 seconds sends them to the parent process through dRB. Later, dRB sends 1 heartbeat for each worker.

### DIFF
--- a/test/aikido/zen/agent_test.rb
+++ b/test/aikido/zen/agent_test.rb
@@ -235,9 +235,9 @@ class Aikido::Zen::AgentTest < ActiveSupport::TestCase
   end
 
   test "#send_heartbeat reports a heartbeat event and updates the settings" do
-    hb = {dummy: :heartbeat}
-    @collector.push_heartbeat(hb)
-    @api_client.expect :report, {"receivedAnyStats" => true}, [hb]
+    dummy_heartbeat = Object.new
+    @collector.push_heartbeat(dummy_heartbeat)
+    @api_client.expect :report, {"receivedAnyStats" => true}, [dummy_heartbeat]
 
     assert_changes -> { Aikido::Zen.runtime_settings.received_any_stats }, to: true do
       @agent.send_heartbeat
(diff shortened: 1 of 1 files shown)


lib/aikido/zen/background_worker.rb
```diff
@@ -0,0 +1,52 @@
+module Aikido::Zen
+  # Generic background worker class backed by queue. Mean to be used by any
```
bitterpanda63:
```suggestion
  # Generic background worker class backed by queue. Meant to be used by any
```

lib/aikido/zen/agent.rb
```diff
@@ -143,11 +143,11 @@ def report(event)
     def send_heartbeat(at: Time.now.utc)
       return unless @api_client.can_make_requests?
 
-      event = @collector.flush(at: at)
-
-      report(event) do |response|
-        updated_settings! if Aikido::Zen.runtime_settings.update_from_json(response)
-        @config.logger.info "Updated runtime settings after heartbeat"
+      @collector.flush_heartbeats.each do |heartbeat|
+        report(heartbeat) do |response|
```
bitterpanda63, line 147: This should only be 1 heartbeat every 10 minutes right? the idea is that we merge all statistics on their server (so in the background process)

lib/aikido/zen/detached_agent/front_object.rb
```diff
@@ -0,0 +1,17 @@
...
+# processes.
+# Every method is called from the child but it runs in the parent process.
+module Aikido::Zen::DetachedAgent
+  class FrontObject
+    def initialize(config: Aikido::Zen.config, collector: Aikido::Zen.collector)
+      @config = config
+      @collector = collector
+    end
+
+    def send_heartbeat_to_parent_process(heartbeat)
```
bitterpanda63, line 13: Yeah not really a fan of this, can be something similar, but would not call it a heartbeat since we don't want to report 12 heartbeats to our server when 12 cores are running. We can group data and send it grouped to the parent process where the parent process then parses it and stores it internally?

bitterpanda63: I might be misinterpreting some parts of this code, maybe the heartbeat here means something different than the heartbeat we send in core

lib/aikido/zen/agent.rb
HiDAl: yes. child process will send N heartbeats each 10s (to be kept in the parent process) and each 10m they will be send to servers. The merging process has to be coded though. It's not complex, but I didn't have time.

lib/aikido/zen/detached_agent/front_object.rb
HiDAl: It's the same heartbeat sent to core, the grouping can happen *in the parent process*, but I said in the previous comment, has to be coded.

PR #95, opened 2025-04-01
Fix rate limiting and user blocking to work with the dRB change

HiDAl:


lib/aikido/zen/detached_agent/front_object.rb
```diff
@@ -5,13 +5,33 @@
...
       @collector.push_heartbeat(heartbeat)
     end
+
+    # Method called by child processes to get an up-to-date version of the
+    # runtime_settings
+    def updated_settings
+      @runtime_settings
+    end
+
+    def calculate_rate_limits(route, ip, actor_hash)
```
bitterpanda63, line 32: would maybe rename to should_rate_limit? calculate rate limits sounds misleading, not really calculating anything right, more or less just incrementing bucket and checking if full

HiDAl: It's actually calculating the rate limit state and returning an object with that info. The info is later used [here](https://github.com/AikidoSec/firewall-ruby/pull/95/files#diff-9b7728c1cb3996a4c04df0d7308d3321da5360c0abeff97116772ed453de2465R48-R52) to decorate the env

PR #106, opened 2025-05-30
Only count requests that should be tracked

timokoessler:


lib/aikido/zen/middleware/request_tracker.rb
```diff
@@ -13,14 +13,16 @@ def call(env)
...
-        if Aikido::Zen.config.collect_api_schema? && request.route && track?(
+        if track?(
           status_code: response[0],
           route: request.route.path,
           http_method: request.request_method
         )
-          Aikido::Zen.track_discovered_route(request)
+          Aikido::Zen.track_request request
+
+          if Aikido::Zen.config.collect_api_schema? && request.route
```
hansott: Do we need to check `collect_api_schema`? We still want to count requests even if API spec is disabled?

timokoessler: Yep, this is done in the line `Aikido::Zen.track_request request` above

hansott: Ah yes, I was confused for a moment!

PR #118, opened 2025-07-07
Make sink methods resilient

marksmith:
This is a significant change; it builds on earlier insights and attempts to ensure, in a systematic way, that `aikido-zen` sinks do not change the behavior of the methods that they replace. To do this it establishes and implements 3 principles:

1) Sink methods should ensure that the original method is called, passing all parameters (positional, keyword, and block) exactly as originally provided
2) Sink methods signatures should not predict/restrict the signature of the original method
3) Unexpected errors that are encountered in sink methods should not interfere with the operation of the original method

tl;dr: sink methods should be transparent, even when inoperable!

To assist with writing sink methods that are reliably transparent, the change introduces meta-methods for defining sink methods that execute code before, after, and around methods. Wherever possible these meta-methods have been used to re-implement the existing sinks; in the few cases where this was not possible, just the divergent sink methods have been conservatively re-implemented to follow the principles above.

The change is big because there are many existing sink methods.

Along the way, numerous violations of the above principles (and-so transparency), inconsistencies, and potential errors were corrected.

It was recognized that the http.rb Gem is not supported or tested on Ruby <3.0 and breaks when the sink methods are called correctly (because of subtle differences in the way that keyword parameters are handled between Ruby versions; particularly between Ruby <2.7, Ruby 2.7, and Ruby >3.0.) As the http.rb project does not support or test the http.rb Gem on Ruby <3.0 I have opted to excluded http.rb on Ruby <3.0, instead of working around the issue encountered to maintain the existing broken behavior.

The change includes a necessary fix to prevent recursion in sink methods by detecting and disabling recursive scanning. This is necessary for safety, as it attempts to ensure that all sink methods return. This is an error class that has been encountered but is not raised; it is a violation of principle 3, and interferes with the operation of the original method. Unknown error classes may exist that remain to be handled.

Please also note that the change explicitly modifies code coverage, out of necessity, for the tests to pass. This is not directly related to the intentional change discussed above.

A case could certainly be made for explicit testing of the `Aikido::Zen::Sinks::DSL`, but this new code was not the cause of the tests failing. The approach taken here was broadly to recognize that many branches do not require a negative-test, and some situations will/can never be tested adequately. Further discussion is invited!

### DIFF
--- a/lib/aikido/zen/sinks_dsl.rb
+++ b/lib/aikido/zen/sinks_dsl.rb
@@ -83,8 +83,8 @@ module Aikido::Zen
         raise
       rescue PresafeError => err
         raise err.cause
-      rescue
-        # empty
+      rescue => err
+        Aikido::Zen.config.logger.debug("[safe] #{err.class}: #{err.message}")
       end
 
       # Presafely execute the given block


lib/aikido/zen/sinks.rb
```diff
@@ -1,6 +1,11 @@
 # frozen_string_literal: true
 
+# Code coverage is disabled in this file because it is environment-specific and
+# not intended to be tested directly.
```
bitterpanda63, line 4:
```suggestion
# not intended to be tested directly. (we do not check with a module installed vs not installed)
```

lib/aikido/zen/sinks_dsl.rb
```diff
@@ -0,0 +1,180 @@
+# frozen_string_literal: true
+
+module Aikido::Zen
+  module Sinks
+    module DSL
+      extend self
+
+      class SafeError < StandardError
```
bitterpanda63: rename safe error

lib/aikido/zen/sinks_dsl.rb
```diff
@@ -0,0 +1,180 @@
+# frozen_string_literal: true
+
```
bitterpanda63, line 2: add global comment explaining unsafe vs safe wrappers

lib/aikido/zen/sinks/httpclient.rb
```diff
@@ -1,68 +1,93 @@
...
+          Scanners::SSRFScanner::Request.new(
             verb: req.http_header.request_method,
             uri: req.http_header.request_uri,
             headers: req.headers
           )
         end
 
         def self.wrap_response(resp)
-          Aikido::Zen::Scanners::SSRFScanner::Response.new(
+          # To understand why code coverage is disabled here see the comment in
```
bitterpanda63: copy comment

lib/aikido/zen/sinks/excon.rb
```diff
@@ -35,33 +26,44 @@ def self.build_request(connection, request)
...
 
-          Aikido::Zen::Scanners::SSRFScanner::Request.new(
+          Scanners::SSRFScanner::Request.new(
             verb: request.fetch(:method) { connection[:method] },
             uri: uri,
             headers: connection[:headers].to_h.merge(request[:headers].to_h)
           )
         end
 
-        def request(params = {}, *)
```
bitterpanda63, line 45: would see if this is removed correctly

lib/aikido/zen/sinks/kernel.rb
```diff
@@ -3,71 +3,32 @@
...
-            command: command,
-            operation: operation
-          )
-        end
+      def self.load_sinks!
+        ::Kernel.singleton_class.prepend(KernelExtensions)
+        ::Kernel.prepend(KernelExtensions)
+      end
 
-        # `system, spawn` functions can be invoked in several ways. For more details,
```
bitterpanda63, line 28: think this might still be useful to explain which functions we wrap and why?

lib/aikido/zen/sinks_dsl.rb
```diff
@@ -0,0 +1,240 @@
...
+module Aikido::Zen
+  module Sinks
+    module DSL
+      extend self
+
+      # In the context of `Aikido::Zen::Sinks::DSL`, the terms safe and presafe
+      # are defined as follows:
+      #
+      # safe: the desired state for a sink, particularly with respect to rescue.
```
bitterpanda63, line 11: let's define as handling unintended errors, while allowing UnderAttackErrors to disrupt the original method?

bitterpanda63: (so just specify the UnderAttackError in the definition, so that it's clear which errors are allowed through)

lib/aikido/zen/sinks_dsl.rb
```diff
@@ -0,0 +1,240 @@
...
+      class PresafeError < StandardError
+      end
+
+      # Safely execute the given block
+      #
+      # All standard errors are suppressed except `Aikido::Zen::UnderAttackError`s.
+      # This ensures that unexpected errors do not interrupt the execution of the
+      # original method, while all detected attacks are raised.
+      #
+      # Error suppression is disabled when `Aikido::Zen.config.debugging?` is true.
```
bitterpanda63: how do you enable that? because our AIKIDO_DEBUG=1 should not cause this, we should still suppress errors

lib/aikido/zen/sinks_dsl.rb
```diff
@@ -0,0 +1,240 @@
...
+          yield
+        else
+          begin
+            yield
+          rescue Aikido::Zen::UnderAttackError
+            raise
+          rescue PresafeError => err
+            raise err.cause
+          rescue
+            # empty
```
bitterpanda63: let's log the error?

lib/aikido/zen/sinks/kernel.rb
marksmith: I would ordinarily agree, but all this comment actually says is that the first argument to `system` or `spawn` may be an optional environment, and that the command-line string given will be executed. These are definitional to `system` and `spawn` and should not need spelling out. The official documentation for these methods does a better job of explaining this.

I'm not sure why the original code made this so difficult. Figuring out why the comment was written and what was really being noted was more work than rewriting the buggy code it commented; we just want to ignore the optional environment so we can scan the command-line. I replaced the comment above with a simpler comment on the simpler code that addresses just this complication.

It should be clear from the context that we are using `ShellInjectionScanner` to scan for shell injection attacks.

lib/aikido/zen/sinks_dsl.rb
marksmith: How would you like to enable it? There was some discussion about what debugging mode was intended for but there needs to be a way to disable error suppression.

lib/aikido/zen/sinks_dsl.rb
marksmith: Perhaps while debugging? Otherwise, the intention was to silently suppress the errors. So far these errors have been transient timing and environmental errors that we don't have much control over and don't intend to fix.

What do you intend to do with any errors that are logged?

lib/aikido/zen/sinks_dsl.rb
marksmith: The comment attached to the `safe` method that is referred throughout this global comment already explains (just a few lines below):

> All standard errors are suppressed except `Aikido::Zen::UnderAttackError`s.

And gives the reasoning:

>  This ensures that unexpected errors do not interrupt the execution of the original method, while all detected attacks are raised.

It is unfortunate that we call UnderAttack an error (when it is in fact a condition implemented as a StandardError) but I think this clearly delineates the what is meant.

I'm wearing of repeating too much in too many places. If you still think it's necessary to give a negative definition (what is not an unintended error) then I can add one, but my feeling is that there are sufficient comments in the DSL module to explain it.

lib/aikido/zen/sinks/kernel.rb
bitterpanda63: yep comment was indeed not very obvious, not necessary with the shift now

lib/aikido/zen/sinks_dsl.rb
bitterpanda63: needs to be debug logged, we do it in all other errors, can happen in customer app, and then we ask for debug logs so we can go ahead and fix the issue

lib/aikido/zen/sinks_dsl.rb
marksmith: Debug mode currently controls maximum log verbosity. For consistency between Zen implementations, for now, it should not be used otherwise.

PR #120, opened 2025-07-16
Improve platform support

marksmith:
This change is necessary to allow running (testing) on Alpine Linux (common in containers), which uses an alternative libc. It allows running on any Linux distribution that uses musl libc.

To achieve this, the many inconsistencies in platform handling needed to be unpicked. The result of addressing these inconsistencies is that support for a platform or specific version can be controlled by adding/removing its platform string (used internally by Ruby/Gems), and its library suffix.

It is now possible to add or extend support to platforms where no pre-built artifact exists; instead of being downloaded, `libzen` may be built locally and copied into `lib/aikido/zen`.

This change adds support for the following platforms (building on my earlier personal effort to `extend-os-support`):

- arm64-linux-musl
- x86_64-linux-musl
- x86_64-freebsd (no pre-built artifact to download)
- x86_64-solaris (no pre-built artifact to download)

### DIFF
--- a/tasklib/libzen.rake
+++ b/tasklib/libzen.rake
@@ -77,11 +77,12 @@ LIBZENS = [
   LibZen.new("arm64-linux.so", "libzen_internals_aarch64-unknown-linux-gnu.so"),
   LibZen.new("arm64-linux-musl.so", "libzen_internals_aarch64-unknown-linux-musl.so"),
   LibZen.new("x86_64-darwin.dylib", "libzen_internals_x86_64-apple-darwin.dylib"),
-  LibZen.new("x86_64-freebsd.so"),
   LibZen.new("x86_64-linux.so", "libzen_internals_x86_64-unknown-linux-gnu.so"),
   LibZen.new("x86_64-linux-musl.so", "libzen_internals_x86_64-unknown-linux-musl.so"),
-  LibZen.new("x86_64-solaris.so"),
-  LibZen.new("x86_64-mingw64.dll", "libzen_internals_x86_64-pc-windows-gnu.dll")
+  LibZen.new("x86_64-mingw64.dll", "libzen_internals_x86_64-pc-windows-gnu.dll"),
+  # Not officially supported, but used during testing:
+  LibZen.new("x86_64-freebsd.so"),
+  LibZen.new("x86_64-solaris.so")
 ].filter(&:resolvable?)
 
 namespace :libzen do


lib/aikido/zen/version.rb
```diff
@@ -5,6 +5,6 @@ module Zen
     VERSION = "1.0.0-beta.1"
 
     # The version of libzen_internals that we build against.
-    LIBZEN_VERSION = "0.1.37"
+    LIBZEN_VERSION = "0.1.41"
```
bitterpanda63:
```suggestion
    LIBZEN_VERSION = "0.1.39"
```

lib/aikido/zen/internals.rb
```diff
@@ -13,14 +13,38 @@ class << self
...
+      lib_ext = FFI::Platform::LIBSUFFIX
+
+      platform = Gem::Platform.local.dup
+
+      names = []
+
+      names << "#{lib_name}-#{platform}.#{lib_ext}"
+
+      unless platform.version.nil?
+        platform.version = nil
```
bitterpanda63, line 36: so if a platform version is defined, we also add the lib but without the platform version? I'm guessing this is to try and get the binary for platforms we don't support yet?

lib/aikido/zen/internals.rb
```diff
@@ -13,14 +13,38 @@ class << self
...
+      end
+
+      names
+    end
+
+    def self.load_libzen
+      libzen_names.each do |libzen_name|
+        libzen_path = File.expand_path(libzen_name, __dir__)
+        begin
+          return ffi_lib(libzen_path)
```
bitterpanda63, line 48: so we test 1 with platform version (first) and second without platform version (2nd) and return once we have a library that works? is there a reason we want to try and test twice, and not just always without platform version?

lib/aikido/zen/internals.rb
```diff
@@ -13,14 +13,38 @@ class << self
...
+        rescue LoadError
+          # empty
+        end
+      end
+      raise LoadError, "Could not load libzen"
```
bitterpanda63, line 53: this is try-catched somewhere? We want an error-level log (so always visible), but we don't want to crash the app if we can't load in the zen binary

tasklib/libzen.rake
```diff
@@ -1,128 +1,134 @@
...
-]
+LIBZENS = [
+  LibZen.new("arm64-darwin.dylib", "libzen_internals_aarch64-apple-darwin.dylib"),
+  LibZen.new("arm64-linux.so", "libzen_internals_aarch64-unknown-linux-gnu.so"),
+  LibZen.new("arm64-linux-musl.so", "libzen_internals_aarch64-unknown-linux-musl.so"),
+  LibZen.new("x86_64-darwin.dylib", "libzen_internals_x86_64-apple-darwin.dylib"),
+  LibZen.new("x86_64-freebsd.so"),
+  LibZen.new("x86_64-linux.so", "libzen_internals_x86_64-unknown-linux-gnu.so"),
+  LibZen.new("x86_64-linux-musl.so", "libzen_internals_x86_64-unknown-linux-musl.so"),
+  LibZen.new("x86_64-solaris.so"),
```
bitterpanda63: these can be removed since they aren't supported?

lib/aikido/zen/internals.rb
marksmith: The way that Ruby represents platforms is as a triplet: `cpu`, `os`, and `version`.

What `version` is used for depends on the `os` i.e. on Linux `version` is used for `musl`, but on macOS `version` is the kernel version number.

Here we are loosely following the Gem loading behavior: If the library (`libzen`) exists for a specific platform version then we attempt to load it, otherwise we fall back and attempt to load the generic version.

This allows us to provide `version` specific builds of `libzen` (which we use for musl), but the real reason for doing this is logical consistency, without which we have to handle all of the special cases.

lib/aikido/zen/internals.rb
marksmith: Correct, that's the goal.

The short answer is: to handle platforms consistently.

See my previous comment:

https://github.com/AikidoSec/firewall-ruby/pull/120#discussion_r2210334208

The longer answer is:

On Linux we always load `libzen` on the first attempt, because `Gem::Platform.local` is `cpu-linux` if the Linux distribution uses glibc, and `cpu-linux-musl` if it uses musl. On macOS we expect to load `libzen` on the second attempt because `Gem::Platform.local` always includes the kernel version number. Each platform is different.

This approach was chosen because it should work correctly on any platform.

tasklib/libzen.rake
marksmith: I often use FreeBSD and illumos in development, for DTrace, and will do so again soon. That is why I created the `extend-os-support` branch, which preceded this work.

lib/aikido/zen/internals.rb
marksmith: This is an editing error on my part; LoadError should still be caught.

tasklib/libzen.rake
bitterpanda63: ok but they aren't downlaoded bc they arent "downloadable" (no artifact name), would just add a comment for the ones that are not downloadable

lib/aikido/zen/internals.rb
bitterpanda63: add as comment :)

lib/aikido/zen/internals.rb
```diff
@@ -13,14 +13,38 @@ class << self
...
+      lib_name = "libzen-v#{LIBZEN_VERSION}"
+      lib_ext = FFI::Platform::LIBSUFFIX
+
+      platform = Gem::Platform.local.dup
+
+      names = []
+
+      names << "#{lib_name}-#{platform}.#{lib_ext}"
+
+      unless platform.version.nil?
```
bitterpanda63, line 35:
```suggestion
      # We want to check both the name with and without ruby's `platform.version`, not because we have different binaries
      # per version but because for linux it's used for `musl`. For Mac OS, where it does store the kernel version, we also add the name here, without the version.
      unless platform.version.nil?
```

tasklib/libzen.rake
```diff
@@ -1,128 +1,130 @@
...
```
bitterpanda63:
```suggestion
  LibZen.new("x86_64-solaris.so"), # not officially supported, but used during testing
```

tasklib/libzen.rake
```diff
@@ -1,128 +1,130 @@
...
-  LibZenDL.new(:linux, "aarch64", "libzen_internals_aarch64-unknown-linux-gnu.so"),
-  LibZenDL.new(:linux, "x86_64", "libzen_internals_x86_64-unknown-linux-gnu.so"),
-  LibZenDL.new(:windows, "x86_64", "libzen_internals_x86_64-pc-windows-gnu.dll")
-]
+LIBZENS = [
+  LibZen.new("arm64-darwin.dylib", "libzen_internals_aarch64-apple-darwin.dylib"),
+  LibZen.new("arm64-linux.so", "libzen_internals_aarch64-unknown-linux-gnu.so"),
+  LibZen.new("arm64-linux-musl.so", "libzen_internals_aarch64-unknown-linux-musl.so"),
+  LibZen.new("x86_64-darwin.dylib", "libzen_internals_x86_64-apple-darwin.dylib"),
+  LibZen.new("x86_64-freebsd.so"),
```
bitterpanda63:
```suggestion
  LibZen.new("x86_64-freebsd.so"), # not officially supported, but used during testing
```

tasklib/libzen.rake
bitterpanda63: done

PR #137, opened 2025-08-01
Supported libraries and frameworks version requirement constraints

marksmith:
`README.md` lists the supported libraries and frameworks. This change codifies these constraints.

Sinks are only loaded when requirements for the sinks are met.

`RailsEngine` should only be defined for and `aikido-zen` should only protect `rails >= 7.0`. This is intended to prevent sinks from loading and the agent starting in earlier versions of rails, and therefore make requiring `aikido-zen` a safe "no-op". A warning should be printed on `stdout` when in debugging mode.

NOTE: `Aikido::Zen.config.debugging` may be set using the `AIKIDO_DEBUG` environment variable or in Ruby but it must be set to true before `Aikido::Zen.protect!` is called to show this warning.

### DIFF
--- a/lib/aikido/zen.rb
+++ b/lib/aikido/zen.rb
@@ -49,12 +49,14 @@ module Aikido
     end
 
     # @!visibility private
-    # Returns whether the loaded named Gem specification satisfies the
-    # listed requirements.
+    # Returns whether the loaded gem specification satisfies the listed requirements.
     #
-    # False is returned when the named Gem specification is not loaded.
+    # Returns false if the gem specification is not loaded.
     #
-    # @return [Boolean]
+    # @param name [String] the gem name
+    # @param requirements [Array<String>] a variable number of gem requirement strings
+    #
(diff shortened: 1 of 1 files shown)


lib/aikido/zen/rails_engine.rb
```diff
@@ -1,5 +1,10 @@
 # frozen_string_literal: true
 
+unless Aikido::Zen.satisfy "rails", ">= 7.0"
+  warn "rails < 7.0 is not supported!" if Aikido::Zen.config.debugging?
```
hansott: Is this also with tagging? I guess not? Otherwise include that it's a message coming from Zen?

```suggestion
  warn "aikido: rails < 7.0 is not supported!" if Aikido::Zen.config.debugging?
```

Perhaps we can also remove the debugging flag (I proposed it earlier I know)
Zen won't run on a rails app that's too old, is there even a point of installing Zen then?

```suggestion
  warn "aikido: rails < 7.0 is not supported!"
```

PR #141, opened 2025-08-11
Redesign initialization

marksmith:
This change redesigns initialization, codifying the idea of a source. 

Sinks are no longer loaded in the `Aikido::Zen::RailsEngine` source. Sinks are loaded only if at least one source is loaded.

The no supported libraries or frameworks warning is now considered after the attempt to load the sinks has been made.

The Aikido Agent is [still] started in the source only once the application has been fully initialized.

### DIFF
--- a/lib/aikido/zen/rails_engine.rb
+++ b/lib/aikido/zen/rails_engine.rb
@@ -10,8 +10,6 @@ module Aikido::Zen
     end
 
     initializer "aikido.add_middleware" do |app|
-      next unless config.zen.protect?
-
       app.middleware.use Aikido::Zen::Middleware::SetContext
       app.middleware.use Aikido::Zen::Middleware::CheckAllowedAddresses
       # Request Tracker stats do not consider failed request or 40x, so the middleware


lib/aikido/zen.rb
```diff
@@ -29,23 +26,24 @@ module Zen
...
-        require_relative "zen/rails_engine"
+      unless load_sources! && load_sinks!
+        config.logger.warn "Zen could not find any supported libraries or frameworks. Visit https://github.com/AikidoSec/firewall-ruby for more information."
+        return
       end
 
-      if Aikido::Zen::Sinks.registry.empty?
-        warn "Zen could not find any supported libraries or frameworks. Visit https://github.com/AikidoSec/firewall-ruby for more information."
-      end
+      middleware_installed!
```
timokoessler, line 46: The middleware is installed automatically?

marksmith: `Aikido::Zen.middleware_installed!` marks _the Zen middleware_ as properly installed. 

It's important to note that what we mean by installed middleware is not necessarily traditional middleware and may involved inserting code at a lower-level (monkey patching) or higher-level (framework specific mechanisms) to best implement the features of _the Zen middleware_.

The individual [Rack] middleware are still automatically installed by the Rails source (`Aikido::Zen::RailsEngine`).

PR #143, opened 2025-08-18
Implement an aliasless method chain patch strategy

marksmith:
This change implements an aliasless method chain patch strategy, designed specifically for the Aikido Zen sinks methods. This strategy is maximally transparent and safe in the presence of arbitrary third-party patch strategies provided that this patch strategy is applied first.

This method closes-over the original method rather than creating an alias and thus does not suffer from alias pollution, which can cause conflicts when multiple third-party gems create aliases with the same name. 

To ensure that Aikido Zen patches before other third-party gems, `aikido-zen` should be required after the Bundler has been setup and before any remaining bundled gems are required. `Aikido::Zen.protect!` must be called or `aikido-zen`, while required, will remain inert. 

```
# boot.rb
require 'bundler/setup'
...

# config/application.rb
...
require 'aikido-zen'
Aikido::Zen.protect!

# Require the gems listed in Gemfile, including any gems
# you've limited to :test, :development, or :production.
Bundler.require(*Rails.groups)
...
```

While it is possible to ensure that Aikido Zen will patch first, following the above procedure, it is not possible to guarantee that Aikido Zen will patch last, as required third-party gems may defer patching until an indeterminate point in execution, i.e. after the application has initialized, along with other gems and under external control.

This causes difficult to predict and debug errors when applications require gems that internally use conflicting strategies -- if method chain patches are applied over module prepend patches, infinite recursion occurs when chain method patches call aliased prepend module patches and the prepended patch super-calls the chained patch.

It is not possible to either dictate the patching strategy that third party gems use or to prevent different patch strategies from conflicting, and the various complex methods and heuristics that have been developed to do so are known to have had limited success. These problems can be sidestepped, however, by using an aliasless method chain patch strategy first, rendering the Aikido Zen sink method patches transparent.

See #118 for more details on how this level of transparency has been achieved in `aikido-zen`.

### DIFF
--- a/lib/aikido/zen.rb
+++ b/lib/aikido/zen.rb
@@ -39,7 +39,7 @@ module Aikido
       return unless config.protect?
 
       unless load_sources! && load_sinks!
-        config.logger.warn "Zen could not find any supported libraries or frameworks. Visit https://github.com/AikidoSec/firewall-ruby for more information."
+        config.logger.warn("Zen could not find any supported libraries or frameworks. Visit https://github.com/AikidoSec/firewall-ruby for more information.")
         return
       end
 


lib/aikido/zen/sinks_dsl.rb
```diff
@@ -145,8 +147,10 @@ def sink_before(method_name, &block)
       #
       # @return [void]
       def presafe_sink_after(method_name, &block)
+        original = instance_method(method_name)
```
hansott, line 152: Might throw? https://github.com/AikidoSec/firewall-node/blob/main/library/agent/hooks/wrapExport.ts#L127

The sink instrumentation will likely target more method names to cover them all? While older versions might miss certain methods, or deprecated methods are removed in newer versions?

lib/aikido/zen/sinks_dsl.rb
```diff
@@ -177,18 +181,20 @@ def sink_after(method_name, &block)
...
       # @param method_name [Symbol, String] the name of the method to define
       # @yield the block to execute around the original method
-      # @yieldparam super_call [Proc] the proc that calls the original method
+      # @yieldparam original_call [Proc] the proc that calls the original method
       # @yieldparam args [Array] the positional arguments passed to the original method
       # @yieldparam kwargs [Hash] the keyword arguments passed to the original method
       #
       # @return [void]
       def presafe_sink_around(method_name, &block)
+        original = instance_method(method_name)
```
hansott, line 194: Same

lib/aikido/zen/sinks_dsl.rb
```diff
@@ -109,9 +109,11 @@ def presafe
       #
       # @return [void]
       def presafe_sink_before(method_name, &block)
+        original = instance_method(method_name)
```
hansott, line 112: Same

PR #145, opened 2025-08-21
Alias set_user to track_user

marksmith:
Users of other Zen implementations may expect to call `set_user` to set up user tracking. This change introduces `set_user` as an alias for `track_user`, and proposes to keep `track_user` for internal consistency; the method name prefix `track_` is used consistently to track e.g. requests, redirects, attacks, etc.

### DIFF
--- a/lib/aikido/zen.rb
+++ b/lib/aikido/zen.rb
@@ -165,6 +165,11 @@ module Aikido
       end
     end
 
+    # Align with other Zen implementations, while keeping internal consistency.
+    class << self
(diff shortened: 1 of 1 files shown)


lib/aikido/zen.rb
```diff
@@ -165,6 +165,11 @@ def self.track_user(user)
       end
     end
 
+    # Align with other Zen implementations, while keeping internal consistency.
+    class << self
+      alias_method :set_user, :track_user
```
hansott, line 170: Do we need to update docs?

PR #146, opened 2025-08-21
Update Rails installation instructions

marksmith:


docs/rails.md
```diff
@@ -2,10 +2,28 @@
...
+require_relative "boot"
+
+require "rails/all"
+
+require "aikido-zen"
+Aikido::Zen.protect!
+
+# Require the gems listed in Gemfile, including any gems
+# you've limited to :test, :development, or :production.
+Bundler.require(*Rails.groups)
```
hansott, line 22: Let's add `set_user` example? (See https://github.com/AikidoSec/firewall-node/blob/main/docs/express.md)

bitterpanda63: +1

PR #148, opened 2025-08-25
Add template for future e2e tests

marksmith:
Adding e2e tests is complicated by the need to support many different versions of Ruby. The change adds a template for future e2e tests base on the `rails7.1_path_traversal` sample app, which was itself based on  `rails7.1_sql_injection`. Creating a sample app from scratch remains to be achieved. A best-effort attempt has been made to ensure that this sample app template is clean.


sample_apps/rails7.1_path_traversal/config/application.rb
```diff
@@ -6,13 +6,14 @@
 require "action_view/railtie"
 require "rails/test_unit/railtie"
 
+# Enable protection after `aikido-zen` has been required.
+require "aikido-zen"
```
timokoessler, line 10: How did that work before without the require? 😄

marksmith: `Bundler.require` required `aikido-zen` automatically, when `Aikido::Zen.protect!` came after `Bundler.require` (assuming that require was not set to false for the `aikido-zen` gem in the Gemfile). But `aikido-zen` must be required manually to be able to enable protection earlier.

PR #149, opened 2025-08-26
Improved detached agent socket handling

marksmith:
It is not possible to guarantee that the application will shut down cleanly and thus that the in use DRb Unix domain socket will be removed. When the application exits abruptly a residual DRb Unix domain socket will remain. This change makes detached agent server startup more reliable by removing this residual DRb Unix domain socket when the server starts.

It first checks whether the Unix domain socket file is in use by opening then closing a connection. If the connection is refused an attempt is made to remove the Unix domain socket. If the connection is accepted, i.e. by another DRb server, the Unix domain socket is in use and DRb will raise an error, preventing multiple detached agent servers from starting in multiple processes using the same `detached_agent_socket_path`.

The `Aikido::Zen::DetachedAgent::Server` has been significantly refactored to align with the agent; removing several implementation oddities.

The `Aikido::Zen.config.detached_agent_socket_path` configuration option is now a path. The `detached_agent_socket_uri` is derived from the detached agent path by prepending `drbunix:`.

### DIFF
--- a/lib/aikido/zen/detached_agent/server.rb
+++ b/lib/aikido/zen/detached_agent/server.rb
@@ -27,11 +27,11 @@ module Aikido::Zen::DetachedAgent
     def start!
       @config.logger.info("Starting DRb Server...")
 
+      # Try to ensure that the DRb service can start if the DRb service did
+      # not stop cleanly.
       begin
-        # Try to connect to the Unix domain socket.
+        # Check whether the Unix domain socket is in use by another process.
         UNIXSocket.new(@socket_path).close
-
-        # Connection successful...
       rescue Errno::ECONNREFUSED
         @config.logger.debug("Removing residual Unix domain socket...")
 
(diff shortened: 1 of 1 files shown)


lib/aikido/zen/detached_agent/server.rb
```diff
@@ -1,41 +1,75 @@
...
+      @config.logger.info("Starting DRb Server...")
+
+      begin
+        # Try to connect to the Unix domain socket.
+        UNIXSocket.new(@socket_path).close
+
+        # Connection successful...
+      rescue Errno::ECONNREFUSED
+        # Remove the residual Unix domain socket.
+        FileUtils.rm_f(@socket_path)
```
bitterpanda63, line 39: also add debug/trace log message here saying that you removed a leftover UDS file

lib/aikido/zen/detached_agent/server.rb
```diff
@@ -1,41 +1,78 @@
...
-
-        @server
+    def start!
+      @config.logger.info("Starting DRb Server...")
+
+      # Try to ensure that the DRb service can start if the DRb service did
+      # not stop cleanly.
+      begin
+        # Check whether the Unix domain socket is in use by another process.
+        UNIXSocket.new(@socket_path).close
```
bitterpanda63, line 34:
```suggestion
        UNIXSocket.new(@socket_path).close
        # Connection successful, so rely on the later error handling to ensure we don't start 2 servers
```

PR #156, opened 2025-09-04
Support custom client IP header

marksmith:
This change allows a custom client IP header to be specified. If specified the value of the header is used as the `client_ip`, otherwise, Rails' application configurable `request.remote_ip` is preferred over Rack's globally configurable `request.ip`.

The custom client IP header can be specified by setting either the `AIKIDO_CLIENT_IP_HEADER` environment variable or the `Aikido::Zen.config.client_ip_header` configuration option.

This implementation leans on Rails and Rack to correctly specify the trusted proxies. Like Rails and Rack, the value of the header provided by the trusted proxy is treated as a string.

### DIFF
--- a/docs/proxy.md
+++ b/docs/proxy.md
@@ -1,0 +1,12 @@
+# Proxy settings
+
+We'll automatically use the `x-forwarded-for` header to determine the client's IP address when behind a proxy.
+
+If you're publicly exposing your server without a load balancer in front of it, you should set the `AIKIDO_TRUST_PROXY` env var to `false` to ensure that the correct IP address is used. Otherwise, someone could potentially spoof their IP address by adding the above header and thus bypassing the rate limiting.
+
+If you need to use a different header to determine the client's IP address, you can set the `AIKIDO_CLIENT_IP_HEADER` environment variable to the name of that header. This will override the default `x-forwarded-for` header.
+
+```bash
+# For Fly.io Platform
+AIKIDO_CLIENT_IP_HEADER=HTTP_FLY_CLIENT_IP bin/rails server
+```
--- a/docs/proxy.md
+++ b/docs/proxy.md
@@ -1,10 +1,8 @@
 # Proxy settings
 
-We'll automatically use the `x-forwarded-for` header to determine the client's IP address when behind a proxy.
+We'll automatically use the `HTTP_X_FORWARDED_FOR` header to determine the client's IP address when behind a trusted proxy.
 
-If you're publicly exposing your server without a load balancer in front of it, you should set the `AIKIDO_TRUST_PROXY` env var to `false` to ensure that the correct IP address is used. Otherwise, someone could potentially spoof their IP address by adding the above header and thus bypassing the rate limiting.
-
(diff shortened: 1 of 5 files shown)


docs/proxy.md
```diff
@@ -0,0 +1,12 @@
+# Proxy settings
+
+We'll automatically use the `x-forwarded-for` header to determine the client's IP address when behind a proxy.
+
+If you're publicly exposing your server without a load balancer in front of it, you should set the `AIKIDO_TRUST_PROXY` env var to `false` to ensure that the correct IP address is used. Otherwise, someone could potentially spoof their IP address by adding the above header and thus bypassing the rate limiting.
```
hansott: We'll need to remove `AIKIDO_TRUST_PROXY`? ;)

docs/proxy.md
```diff
@@ -0,0 +1,12 @@
+# Proxy settings
+
+We'll automatically use the `x-forwarded-for` header to determine the client's IP address when behind a proxy.
```
hansott: Let's point to Rails/Rack docs to configure this?

lib/aikido/zen/request.rb
```diff
@@ -40,6 +41,17 @@ def schema
       @schema ||= Aikido::Zen::Request::Schema.build
     end
 
+    # @api private
+    #
+    # @return [String] the IP address of the client making the request.
+    def client_ip
+      return @client_ip if @client_ip
+
+      @client_ip = env[@config.client_ip_header] if @config.client_ip_header
```
hansott:
```suggestion
      @client_ip = env[@config.client_ip_header] if @config.client_ip_header and env[@config.client_ip_header] =~ Resolv::AddressRegex
```

Validate if this is an IP address?

hansott: We decided not to fail silently here.

docs/proxy.md
marksmith: The documentation for this is auto-generated from the Ruby `Rack::Request` and `ActionDispatch::RemoteIp` class definitions and is not well-suited for inclusion here. If instructional or explanatory content is needed, we would need to create it ourselves.

docs/proxy.md
```diff
@@ -0,0 +1,10 @@
+# Proxy settings
+
+We'll automatically use the `HTTP_X_FORWARDED_FOR` header to determine the client's IP address when behind a trusted proxy.
```
hansott, line 3:
```suggestion
We'll automatically use the `x-forwarded-for` header to determine the client's IP address when behind a trusted proxy.
```

docs/proxy.md
```diff
@@ -0,0 +1,10 @@
...
+
+If you need to use a different header to determine the client's IP address, you can set the `AIKIDO_CLIENT_IP_HEADER` environment variable to the name of that header. This will override the default `HTTP_X_FORWARDED_FOR` header.
```
hansott, line 5:
```suggestion
If you need to use a different header to determine the client's IP address, you can set the `AIKIDO_CLIENT_IP_HEADER` environment variable to the name of that header. This will override the default `x-forwarded-for` header.
```

PR #168, opened 2025-09-24
Implement method hardening for File.join

marksmith:
This change adds method hardening, currently only for `File.join`. This is disabled by default and enabled by setting the `Aikido::Zen.config.harden` configuration option or `AIKIDO_HARDEN` environment variable to true.

See #155  for background on why hardening `File.join` is desirable.

The hardened implementation of `File.join` ensures that `File.join` is used as documented. This may be relaxed in future commits to balance safety and compatibility with widespread safe misuses of `File.join`. For this discussion, the exact approach to hardening is less important than the question: _can we fix dangerous undocumented behavior, like that of `File.join`, when it is discovered?_

Internal discussion suggests yes.

### DIFF
--- a/lib/aikido/zen/sinks/file.rb
+++ b/lib/aikido/zen/sinks/file.rb
@@ -108,7 +108,7 @@ module Aikido::Zen
             # Calling File.join with a single string argument returns the string
             # argument itself, having no practical effect. Therefore, it can be
             # presumed that if File.join is called with a single array argument
-            # then this was its intended usage, and the array did not origiate
+            # then this was its intended usage, and the array did not originate
             # from user input that was assumed to be a string.
             strings = args
             strings = args.first if args.size == 1 && args.first.is_a?(Array)


marksmith:
**This PR is at prototype stage and was created for discussion.**

marksmith:
As the subsequent commit message explains, method hardening for `File.join` is no longer optional because the previous solution has been removed. The hardened method has been softened to allow a single non-nested string array argument. A comment explaining the reasoning behind this is included in the definition of `File.join`.

This PR should be considered for merging after #155, so that the final optimized solution is recorded in the Git history.

lib/aikido/zen/sinks/file.rb
```diff
@@ -81,8 +81,46 @@ def self.load_sinks!
...
+            # i.e.
+            #
+            # File.join(["prefix", "filename"])
+            #
+            # This is considered acceptable.
+            #
+            # Calling File.join with a single string argument returns the string
+            # argument itself, having no practical effect. Therefore, it can be
+            # presumed that if File.join is called with a single array argument
+            # then this was its intended usage, and the array did not origiate
```
hansott:
```suggestion
            # then this was its intended usage, and the array did not originate
```

PR #169, opened 2025-10-01
Add configurable initial heartbeats

marksmith:
Two initial heartbeats are configured for 30 seconds and 2 minutes.

Initial heartbeats are implemented separately from the recurring 10-minute heartbeats.

The effect of this change is to send heartbeats at:

```
= 30s
==== 2m
==================== 10m
...
```

Assuming the current 10-minute heartbeat interval.

### DIFF
--- a/lib/aikido/zen/agent.rb
+++ b/lib/aikido/zen/agent.rb
@@ -37,22 +37,22 @@ module Aikido::Zen
     end
 
     def start!
-      @config.logger.info "Starting Aikido agent v#{Aikido::Zen::VERSION}"
+      @config.logger.info("Starting Aikido agent v#{Aikido::Zen::VERSION}")
 
       raise Aikido::ZenError, "Aikido Agent already started!" if started?
       @started_at = Time.now.utc
       @collector.start(at: @started_at)
 
       if @config.blocking_mode?
-        @config.logger.info "Requests identified as attacks will be blocked"
+        @config.logger.info("Requests identified as attacks will be blocked")
       else
-        @config.logger.warn "Non-blocking mode enabled! No requests will be blocked."
+        @config.logger.warn("Non-blocking mode enabled! No requests will be blocked.")
       end
 
       if @api_client.can_make_requests?
-        @config.logger.info "API Token set! Reporting has been enabled."
+        @config.logger.info("API Token set! Reporting has been enabled.")
       else
-        @config.logger.warn "No API Token set! Reporting has been disabled."
+        @config.logger.warn("No API Token set! Reporting has been disabled.")
         return
       end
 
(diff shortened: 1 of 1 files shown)


lib/aikido/zen/agent.rb
```diff
@@ -67,8 +67,11 @@ def start!
 
       poll_for_setting_updates
 
-      @worker.delay(@config.initial_heartbeat_delay) do
-        send_heartbeat if @collector.stats.any?
+      @config.initial_heartbeat_delays.each do |heartbeat_delay|
+        @worker.delay(heartbeat_delay) do
+          @config.logger.info "Executing initial heartbeat after #{heartbeat_delay} seconds"
```
timokoessler: Not sure what the default log level is, but maybe better a debug (if info is logged by default)?

lib/aikido/zen/agent/heartbeats_manager.rb
```diff
@@ -30,11 +30,11 @@ def stale_settings?
     def start(&task)
       return if running?
 
-      if @settings.heartbeat_interval&.nonzero?
```
timokoessler, line 33: I can't really follow this change, isn't it still called `heartbeat_interval` in `runtime_settings.rb`?

lib/aikido/zen/agent.rb
marksmith: `info` is used for similar log messages throughout this file, so I'm inclined to leave this as `info`.

lib/aikido/zen/agent/heartbeats_manager.rb
marksmith: This simply calls an existing method, heartbeat `interval`, which I intended to repurposed to compute the appropriate heartbeat interval based on a heartbeat counter.

https://github.com/AikidoSec/firewall-ruby/blob/69cb14b08480ac84718366676c44c6c392b1a759/lib/aikido/zen/agent/heartbeats_manager.rb#L59-L64

I didn't end up doing this because Zen Ruby already included logic for an initial heartbeat and that made it simpler to extend that logic rather than replacing it with a computed heartbeat interval using a heartbeat counter.

https://github.com/AikidoSec/firewall-ruby/blob/69cb14b08480ac84718366676c44c6c392b1a759/lib/aikido/zen/agent.rb#L70-L76

hansott (approved):

lib/aikido/zen/agent/heartbeats_manager.rb
timokoessler: I just missed this alias at the end of the file. :D

```rb
 def interval
      @settings.heartbeat_interval
    end
```

PR #173, opened 2025-10-22
Fix the k6 benchmark and make it more realistic

marksmith:
This change fixes the k6 benchmark, which starts the `rails7.1_benchmark` (`rails7.1_sql_injection`) sample app with (on port 3001) and without (on port 3002) `firewall-ruby` enabled.

The `test_get_page_with_sql_injection` test has been removed because we have chosen to refocus on normal operation.

The number of virtual users is one and the benchmark now runs for a duration of 60 seconds. Realistic headers are used.

The number of virtual users was initially increased to two, however, this was reverted due to a non-zero `http_req_failed` threshold rate (only on GitHub) when the database is `sqlite3`. (Removing the `http_req_failed` threshold rate check was considered but has been rejected.)

A new `GET /cats/count` endpoint has been introduced for use in the benchmark, which replaces the `GET /cats` endpoint. Rather than returning all of the cats rendered as HTML or JSON, `GET /cats/count` endpoint returns the number of cats rendered as HTML or JSON. The response therefore remains effectively constant in size, despite the continually growing number of cats. Therefore, each iteration should take effectively the same amount of time.

To fix the benchmark, which broke because of a confusion between configuration options set by environment variables in and in Ruby, I have chosen to remove the internal protect check. `Aikido::Zen.protect!` can be called conditionally where protection is desired.

### DIFF
--- a/benchmarks/rails7.1_sql_injection.js
+++ b/benchmarks/rails7.1_sql_injection.js
@@ -94,7 +94,7 @@ export default function () {
   )
 
   test("test_get_page_without_attack",
-    (http) => http.get("/cats",
+    (http) => http.get("/cats/count",
     {
         headers: headers
     })


benchmarks/rails7.1_benchmark.js
```diff
@@ -0,0 +1 @@
+rails7.1_sql_injection.js
```
hansott, line 1: What's the purpose of this file?

lib/aikido/zen/config.rb
```diff
@@ -8,12 +8,6 @@
 
 module Aikido::Zen
   class Config
-    # @api private
-    # @return [Boolean] whether Aikido should protect.
-    def protect?
-      !api_token.nil? || blocking_mode? || debugging?
-    end
```
hansott, line 15: Hmm, what are the implications of removing this?

Protecting will always be enabled when `Aikido::Zen.protect!` is called?

Even if `AIKIDO_TOKEN`, `AIKIDO_BLOCK` and `AIKIDO_DEBUG` isn't set?

What's the purpose of instrumenting an app if we won't even block or reporting anything?

benchmarks/rails7.1_benchmark.js
marksmith: The `rails7.1_benchmark` sample app is a symlink to the `rails7.1_sql_injection` sample app. The example from `benchmarks/README.md` runs the `wrk` benchmark for the `rails7.1_benchmark` sample app.

This file is a symlink, which allows you to also run the `k6` benchmark for the `rails7.1_benchmark` sample app.

Without this symlink, you get an error from `k6` stating that `rails7.1_benchmark.js` does not exist (which is not immediately obvious and requires you to read the `Rakefile` to understand what is happening).

lib/aikido/zen/config.rb
marksmith: If you call `Aikido::Zen.protect!` then protection is enabled, unless explicitly disabled i.e. in the environment by setting the `AIKIDO_DISABLED` environment variable to true.

The additional `protect?` check was added to make it less likely that protection is enabled in supporting processes, which also load `application.rb`. This was a convenience, but leads to confusing issues when configuration options are configured in Ruby, as explained in the rails docs, and done in the sample apps. In this case it subtly broke the `rails7.1_sql_injection` sample app tests and `k6` benchmark.

Since it is possible to control where protection is enabled by conditionally calling `Aikido::Zen.protect!` or setting the `AIKIDO_DISABLED` environment variable to true, and because of subtle issues like the ones described above, I currently feel it is better that this heuristic be removed.

timokoessler: This is somehow a dangerous change for existing users? I also agree that it would be nice to keep this behavior the same across all agents.

marksmith: If I recall correctly, this heuristic is only implemented in `firewall-node` and `firewall-ruby`. `firewall-ruby` has more flexibility in configuration than any of the other agents. This is a mixed blessing, but the difference is big enough that implementing the heuristic causes issues like above, and does not negate the need to carefully consider what processes protection is enabled in, because, unless care is taken, environment variables may also be visible in supporting processes.

So this heuristic (which was not intended for this purpose) is useful when it applies but not without faults and clashes with features that are specific to `firewall-ruby`.

hansott:
```rb
# If you don't want to depend on ENV vars, this is another way to
# configure the Zen library.
if Rails.application.config.respond_to?(:zen)
  Rails.application.config.zen.blocking_mode = true
  Rails.application.config.zen.logger = ::Rails.logger.tagged("aikido")
end
```
^ in a ruby app you can add `config/initializers/zen.rb`

The problem is that this file isn't loaded if `AIKIDO_TOKEN`, `AIKIDO_DEBUG` and `AIKIDO_BLOCK` is missing

We do check `AIKIDO_DISABLE` when `Aikido::Zen.protect!` is called in `application.rb`

So protect! will always instrument code unless `AIKIDO_DISABLE=true`

hansott: Okay to proceed!

PR #198, opened 2025-11-18
Attack wave protection

marksmith:
This change adds attack wave protection. An attack wave is a series of "suspicious" requests, within a defined time frame. The `Aikido::Zen::AttackWave::Detector` currently detects:

- Requests with "suspicious" methods
- Requests with "suspicious" paths, containing suspicious directory names, file names, or file extensions.
- Requests with "suspicious" query parameters, containing SQL keywords.

The suspicious elements are modeled after requests typically made by web scanners.

Attack wave protection can be configured using the `Aikido::Zen.config` configuration options:

- `attack_wave_threshold`: the number of "suspicious" requests required to trigger attack wave detection
- `attack_wave_min_time_between_requests`: the minimum time between requests in milliseconds
- `attack_wave_min_time_between_events`: the minimum time between reports in milliseconds
- `attack_wave_max_cache_entries`: the maximum number of cache entries

This implementation is based on this firewall-node [PR](https://github.com/AikidoSec/firewall-node/pull/706).

### DIFF
--- a/test/aikido/zen/collector/stats_test.rb
+++ b/test/aikido/zen/collector/stats_test.rb
@@ -176,6 +176,10 @@ class Aikido::Zen::Collector::StatsTest < ActiveSupport::TestCase
         attacksDetected: {
           total: 0,
           blocked: 0
+        },
+        attackWaves: {
+          total: 0,
+          blocked: 0
         }
       }
     }
--- a/test/aikido/zen/collector_test.rb
+++ b/test/aikido/zen/collector_test.rb
@@ -201,11 +201,11 @@ class Aikido::Zen::CollectorTest < ActiveSupport::TestCase
           attacksDetected: {
             total: 0,
             blocked: 0
+          },
+          attackWaves: {
+            total: 0,
+            blocked: 0
           }
-        },
-        attackWaves: {
-          total: 0,
-          blocked: 0
         }
       },
       users: [],
@@ -240,11 +240,11 @@ class Aikido::Zen::CollectorTest < ActiveSupport::TestCase
           attacksDetected: {
             total: 0,
             blocked: 0
+          },
+          attackWaves: {
+            total: 5,
+            blocked: 2
           }
-        },
-        attackWaves: {
-          total: 5,
-          blocked: 2
         }
       },
       users: [],
@@ -280,11 +280,11 @@ class Aikido::Zen::CollectorTest < ActiveSupport::TestCase
           attacksDetected: {
             total: 0,
             blocked: 0
+          },
+          attackWaves: {
+            total: 0,
+            blocked: 0
           }
-        },
-        attackWaves: {
-          total: 0,
-          blocked: 0
         },
         sinks: {
           "test" => {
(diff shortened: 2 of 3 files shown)


lib/aikido/zen/attack_wave/helpers.rb
```diff
@@ -0,0 +1,457 @@
...
+        path_parts = path.downcase.split("/")
+
+        file_name = path_parts.pop
+
+        if file_name
+          return true if SUSPICIOUS_FILE_NAMES.include?(file_name)
+
+          filename_parts = file_name.split(".")
+
+          extension = filename_parts.pop
```
timokoessler: In Node we only check the extension if the filename includes a dot. I guess you would check the whole filename as extension in this case?

marksmith: This works because `String#split` always returns an `Array` and `Array#pop` returns `nil` if the `Array` empty.

```ruby
irb(main):001> filename_parts = "filename".split(".")
=> ["filename"]
irb(main):002> filename = filename_parts.pop
=> "filename"
irb(main):003> extension = filename_parts.pop
=> nil
irb(main):004> ["env", "bak", "sql"].include?(extension)
=> false
irb(main):005>
```

The first `Array#pop` always returns the filename and the second `Array#pop` either returns the extension or `nil`. Since `SUSPICIOUS_FILE_EXTENSIONS` does not include `nil`, execution continues.

lib/aikido/zen/middleware/attack_wave_protector.rb
```diff
@@ -0,0 +1,37 @@
...
+    module Middleware
+      class AttackWaveProtector
+        def initialize(app, zen: Aikido::Zen)
+          @app = app
+          @zen = zen
+        end
+
+        def call(env)
+          context = @zen.current_context
+          protect(context)
```
PopoviciMarian, line 16: The `AttackWaveProtector` middleware currently runs as pre-middleware (detecting attacks before the controller executes). This means `set_user()` calls in controllers haven't executed yet, so attack wave events are sent without user information.

Fix:
```
def call(env)
  response = @app.call(env)  # Controller runs first
  
  context = @zen.current_context
  protect(context)  # ✓ Runs AFTER controller (user is set)
  
  response
end
```

timokoessler: In Node we are running it when the response is finished, just before closing the request.

lib/aikido/zen/rails_engine.rb
```diff
@@ -17,6 +17,7 @@ class RailsEngine < ::Rails::Engine
       # Request Tracker stats do not consider failed request or 40x, so the middleware
       # must be the last one wrapping the request.
       app.middleware.use Aikido::Zen::Middleware::RequestTracker
+      app.middleware.use Aikido::Zen::Middleware::AttackWaveProtector
```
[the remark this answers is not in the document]
marksmith: The comment states that `RequestTracker` must be "the last one wrapping the request".

PR #211, opened 2025-12-02
Fix socket resource leak when stored SSRF is blocked

PopoviciMarian:
- When a stored SSRF attack is detected and blocked in the socket sink, the TCP socket is not being properly closed, leading to resource leaks. This could cause file descriptor exhaustion.


marksmith:
This PR was split and superseded by PRs #212 and #213.

PR #219, opened 2025-12-18
Improve middleware registration reliability

marksmith:
This change uses the `Rails::Rack::Logger` middleware as a well-known stable anchor point and inserts our middleware as a block directly after this. The `aikido.add_middleware` initializer is now called after any application configuration and initializers are finished; helping to ensure that our middleware is inserted directly after the anchor point and before any middleware registered at that anchor point by the application.

It is not possible to completely control middleware registration. The possibility of (possibly optional) manual middleware registration has been briefly discussed and may be considered again in the future if necessary.

## Summary by Aikido Security
**⚡ Enhancements**
* Improved middleware registration using Rails::Rack::Logger as anchor.

**🔧 Refactors**
* Reordered middleware insertion to run after application initializers.

### DIFF
--- a/lib/aikido/zen/rails_engine.rb
+++ b/lib/aikido/zen/rails_engine.rb
@@ -9,16 +9,24 @@ module Aikido::Zen
       config.zen = Aikido::Zen.config
     end
 
-    initializer "aikido.add_middleware" do |app|
-      app.middleware.insert_before 0, Aikido::Zen::Middleware::ForkDetector
+    initializer "aikido.add_middleware", after: :load_config_initializers do |app|
+      # The midleware to be inserted in order. The first middleware is the existing
+      # middleware to use as an anchor point.
+      middleware = [
+        ::Rails::Rack::Logger,
+        Aikido::Zen::Middleware::ForkDetector,
+        Aikido::Zen::Middleware::ContextSetter,
+        Aikido::Zen::Middleware::AllowedAddressChecker,
+        Aikido::Zen::Middleware::AttackProtector,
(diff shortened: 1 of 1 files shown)


lib/aikido/zen/rails_engine.rb
```diff
@@ -9,16 +9,24 @@ class RailsEngine < ::Rails::Engine
...
-      app.middleware.use Aikido::Zen::Middleware::AttackProtector
-      app.middleware.use Aikido::Zen::Middleware::AttackWaveProtector
-      # Request Tracker stats do not consider failed request or 40x, so the middleware
-      # must be the last one wrapping the request.
-      app.middleware.use Aikido::Zen::Middleware::RequestTracker
+    initializer "aikido.add_middleware", after: :load_config_initializers do |app|
+      # The midleware to be inserted in order. The first middleware is the existing
+      # middleware to use as an anchor point.
+      middleware = [
+        ::Rails::Rack::Logger,
```
timokoessler, line 16: Also add the PR comment info in code, why the logger is needed? Is this middleware always registered already?

marksmith: As the comment notes:

> The first middleware in the existing middleware to use as an anchor point.

`Rails::Rack::Logger` exists and is expected to be registered, however, this cannot truly be guaranteed because the middleware chain can, in principle, be freely modified. In practice, `Rails::Rack::Logger` is registered as part of the framework and is a well-known stable anchor point.

timokoessler (approved):
But indeed better to determine why zen internals is failing :)

PR #260, opened 2026-03-27
Handle invalid JSON body payload sources

marksmith:
This change handles invalid JSON body payload sources, not extracting body payload sources if the JSON-encoded body could not be parsed. This prevents an `ActionDispatch::Http::Parameters::ParseError` from being raised and propagating, causing the `Aikido::Zen::Middleware::RequestTracker` and `Aikido::Zen::Middleware::AttackWaveProtector` middleware to fail after the application has already processed the request.

The susceptibility of other payload sources has been investigated and discussed internally.

## Summary by Aikido
|  Security Issues: 0 |  Quality Issues: 0 |  Resolved Issues: 0 |
| :--- | :--- | :--- |

**🔧 Refactors**
* Rescued exceptions when extracting Rails request payload and metadata
* Wrapped Rack request parameter accesses with rescue to return defaults

<sup>[More info](https://app.aikido.dev/featurebranch/scan/112407245?groupId=6)</sup>

### DIFF
--- a/lib/aikido/zen/context/rails_request.rb
+++ b/lib/aikido/zen/context/rails_request.rb
@@ -34,19 +34,20 @@ module Aikido::Zen
     end
 
     Context.new(request) do |req|
-      body = begin
-        req.request_parameters
-      rescue ActionDispatch::Http::Parameters::ParseError
-        {}
-      end
+      query = req.query_parameters rescue {}
+      body = req.request_parameters rescue {}
+      route = req.path_parameters rescue {}
+      header = req.normalized_headers rescue {}
+      cookie = decrypt_cookies.call(req) rescue {}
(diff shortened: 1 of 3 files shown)


lib/aikido/zen/context/rails_request.rb
```diff
@@ -34,9 +34,15 @@ def self.router
...
+      body = begin
+        req.request_parameters
+      rescue ActionDispatch::Http::Parameters::ParseError
+        {}
+      end
+
       {
         query: req.query_parameters,
-        body: req.request_parameters,
+        body: body,
```
hansott, line 75: maybe safer to put the other calls also in the begin ... rescue block? (req.path_parameters, ...)

hansott (approved):

PR #273, opened 2026-04-20
Optimize IP lists

marksmith:
This change optimizes the implementation of IP lists, used for blocked, allowed, and monitored IP lists, introduced in #254; reverted in #268 due to excessive CPU usage during testing.

IP lists were perviously implemented using the pre-existing `Aikido::Zen::RuntimeSettings::IPSet` class. `IPSet` uses `Set` internally to provide O(1) membership tests, but resorts to `Set#any?` in order to support CIDR ranges. `Set#any?` accepts a predicate block and applies the predicate to each of the set items, making the membership test O(n).

The implementation of `Aikido::Zen::RuntimeSettings::IPList` from `Aikido::Zen::RuntimeSettings::IPSet` are separated so that `IPList`s can be implemented independently of `IPSet`s, to localize the effects of this change. If this optimization is effective for `IPList` then it may be considered for `IPSet`, if the same constraints are applicable.

The optimized implements binary search over IP list ranges and assumes that IP list ranges are unique and non-overlapping, making the membership test O(log n), with minimal complexity. This is considered the most pragmatic optimization choice.

## Summary by Aikido
|  Security Issues: 0 | 🔍 Quality Issues: 3 |  Resolved Issues: 0 |
| :--- | :--- | :--- |

**⚡ Enhancements**
* Optimized IP list membership checks with binary-search IPList implementation
* Integrated IP list support into runtime, agent, middleware, and tests
* Expanded collector to track IP list metrics and events

<sup>[More info](https://app.aikido.dev/featurebranch/scan/106124076?groupId=6)</sup>


test/aikido/zen/runtime_settings_test.rb
```diff
@@ -282,6 +282,116 @@ class Aikido::Zen::RuntimeSettingsTest < ActiveSupport::TestCase
...
+        {"key" => "claude_user", "pattern" => "Claude-User"},
+        {"key" => "adsbot_google", "pattern" => "AdsBot-Google"},
+        {"key" => "mediapartners_google", "pattern" => "Mediapartners-Google"},
+        {"key" => "mediapartners_googlebot", "pattern" => "Mediapartners \\(Googlebot\\)"},
+        {"key" => "google_adwords", "pattern" => "Google-Adwords"}
+      ]
+    })
+
+    assert_kind_of Array, @settings.blocked_ip_lists
+    assert 1, @settings.blocked_ip_lists.size
```
hansott: assert(test, msg = nil)

1 will always be true?

There are more below!

test/aikido/zen/runtime_settings/ip_list_test.rb
```diff
@@ -0,0 +1,94 @@
+# frozen_string_literal: true
+
+require "test_helper"
+
+class Aikido::Zen::RuntimeSettings::DomainsTest < ActiveSupport::TestCase
```
hansott:
```suggestion
class Aikido::Zen::RuntimeSettings::IPListTest < ActiveSupport::TestCase
```

lib/aikido/zen/config.rb
```diff
@@ -322,10 +322,18 @@ def read_boolean_from_env(value)
...
 
     # @!visibility private
-    DEFAULT_BLOCKED_RESPONDER = ->(request, blocking_type) do
+    DEFAULT_BLOCKED_RESPONDER = ->(request, blocking_type, reason = nil) do
       message = case blocking_type
       when :ip
-        format("Your IP address is not allowed to access this resource. (Your IP: %s)", request.ip)
+        "Your IP address is not allowed to access this resource. (Your IP: #{request.ip})"
+      when :ip_allowed_list
+        "Your IP address is not allowed to access this resource. (Your IP: #{request.client_ip})"
```
hansott, line 330: Is it intentional to use client_ip vs ip (see above)?

marksmith: Discussed internally. I will create a follow-up PR if we decide to change `ip` to `client_ip`.

hansott (approved):

PR #280, opened 2026-05-05
Fix header normalization

marksmith:
This change resolves a `NoMethodError` in `Aikido::Zen::Request#normalized_headers`. When a header name contains consecutive hyphens, `part` is `""` and `part[0]` returns `nil`; `nil` does not respond to `upcase`:

> undefined method `upcase' for nil (NoMethodError)

The Rails integration test added to the `rails7.2_generic` application asserts that `aikido-zen` detects SQL injection in a cookie with a poison header name containing consecutive hyphens in an attempt to halt scanning early.

## Summary by Aikido
|  Security Issues: 0 | 🔍 Quality Issues: 1 |  Resolved Issues: 0 |
| :--- | :--- | :--- |

**🐛 Bugfixes**
* Fixed header normalization to avoid NoMethodError on empty parts
* Fixed cached instance variable name when delegate was replaced

<sup>[More info](https://app.aikido.dev/featurebranch/scan/112361533?groupId=6)</sup>

### DIFF
--- a/lib/aikido/zen/request.rb
+++ b/lib/aikido/zen/request.rb
@@ -64,10 +64,14 @@ module Aikido::Zen
     def normalized_headers
       @normalized_headers ||= env.slice(*BLESSED_CGI_HEADERS)
         .merge(env.select { |key, _| key.start_with?("HTTP_") })
-        .transform_keys { |header|
-          name = header.sub(/^HTTP_/, "").downcase
-          name.split("_").map { |part| part[0].upcase + part[1..] }.join("-")
-        }
+        .transform_keys do |header|
+          header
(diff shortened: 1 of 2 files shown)


lib/aikido/zen/request.rb
```diff
@@ -64,10 +64,14 @@ def client_ip
     def normalized_headers
       @normalized_headers ||= env.slice(*BLESSED_CGI_HEADERS)
```
hansott, line 65: unrelated but `__setobj__` resets `@normalized_header` instead of `@normalized_headers`

hansott (approved):

PR #285, opened 2026-05-27
Realtime settings updates

marksmith:
This change adds realtime settings updates from server-side events. The `Aikido::Zen::Agent` attempts to connect the realtime endpoint at startup and warns the user that realtime settings updates will not be available because the startup connection attempt failed.

Realtime settings updates are implemented on `Aikido::Zen::APIStream`. The API stream implements a simple server-side events client, which connects to the realtime endpoint and reads events. The API streams executes concurrently using a `Concurrent::SingleThreatExecutor`, and calls all registered event handlers for each event received. Currently, the agent registers a single event handler for the `config-updated` event.

Event handlers are executed by the single thread executor, and access to the runtime settings object must be synchronized; because of this change, the runtime settings object is now accessed from multiple threads. Synchronization is localized in the agent using the `update_settings_from_runtime_config!` and `update_settings_from_runtime_firewall_lists!` methods, defined using `updater`.

The change includes a number of important tangential fixes. Notable is the fix to outbound connection blocking with multiple processes, where the default `Aikido::Zen::RuntimeSettings::DomainSettings.none` object identity check fails because each process instantiates its own process local default after forking. The globally unique sentinal `nil` is now used as the default, and the implementation and tests for outbound connection blocking have been modified.

## Summary by Aikido
|  Security Issues: 0 | 🔍 Quality Issues: 11 |  Resolved Issues: 0 |
| :--- | :--- | :--- |

**🚀 New Features**
* Added APIStream SSE client and integrated realtime updates into Agent

**⚡ Enhancements**
* Probed realtime endpoint at startup and enabled configurable endpoint usage

**🐛 Bugfixes**
* Fixed timestamp unit handling and return values when parsing runtime settings

**🔧 Refactors**
* Replaced single-thread executor with Thread and improved stream stop logic
* Introduced mutexes in Agent to synchronize runtime settings updates across threads

<sup>[More info](https://app.aikido.dev/featurebranch/scan/134164517?groupId=6)</sup>


test/aikido/zen/api_stream_test.rb
```diff
@@ -0,0 +1,237 @@
+# frozen_string_literal: true
+
+require "test_helper"
+require "securerandom"
+
+class Aikido::Zen::StreamTest < ActiveSupport::TestCase
+  setup do
+    @endpoint = "https://runtime.aikido.dev/api/runtime/stream"
```
timokoessler: The new endpoint is zen.aikido.dev

lib/aikido/zen/agent.rb
```diff
@@ -177,23 +190,88 @@ def send_heartbeat(at: Time.now.utc)
...
+          updated_settings!
+          @config.logger.info("Updated runtime settings after server-side event")
+
+          update_settings_from_runtime_firewall_lists!(@api_client.fetch_runtime_firewall_lists)
+          @config.logger.info("Updated runtime firewall list after server-side event")
+        end
+      end
+    end
+
+    private
```
timokoessler: I guess this private should not be in this line but on the method above or below?

lib/aikido/zen/api_stream.rb
```diff
@@ -0,0 +1,197 @@
...
+      end
+
+      true
+    end
+
+    def stop!
+      return false unless @running.make_false
+
+      @executor.shutdown
+      @executor.wait_for_termination(@read_timeout)
```
timokoessler: Waiting up to 70s on shutdown is too long?

lib/aikido/zen/agent.rb
```diff
@@ -177,23 +190,88 @@ def send_heartbeat(at: Time.now.utc)
...
 
-          Aikido::Zen.runtime_settings.update_from_runtime_firewall_lists_json(@api_client.fetch_runtime_firewall_lists)
+          update_settings_from_runtime_firewall_lists!(@api_client.fetch_runtime_firewall_lists)
           @config.logger.info("Updated runtime firewall list after polling")
         end
       end
     end
 
-    private def heartbeats
+    def settings_updated(event)
```
timokoessler, line 213: Can be private?

lib/aikido/zen/agent.rb
```diff
@@ -177,23 +190,88 @@ def send_heartbeat(at: Time.now.utc)
...
+      updated_at = Time.at(event[:data]["configUpdatedAt"].to_i)
```
timokoessler, line 214: What happens if data is nil when received via SSE?

test/aikido/zen/agent_test.rb
```diff
@@ -87,22 +111,98 @@ def report(event)
...
+    stub_probe_realtime_endpoint
+
     @config.api_token = nil
     @agent.start!
 
     assert_logged :warn, /no api token set! reporting has been disabled/i
     refute_logged :debug, /api token set! reporting has been enabled/i
   end
 
+  test "#start! probes the realtime endpont" do
```
timokoessler: Small typo :)

lib/aikido/zen/api_stream.rb
```diff
@@ -0,0 +1,197 @@
...
+      @port = @config.realtime_endpoint.port
+      @use_ssl = @config.realtime_endpoint.scheme == "https"
+      @token = @config.api_token
+
+      @handlers = Concurrent::Array.new
+    end
+
+    # @return [Boolean] whether we could connect to the realtime endpoint
+    def can_connect?
+      http = Net::HTTP.new(@host, @port)
```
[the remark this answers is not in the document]
marksmith: `Net::HTTP` is configured twice but is configured differently.

lib/aikido/zen/agent.rb
marksmith: There are a few ways to use the private keyword in Ruby. In this usage all subsequent methods are private.

lib/aikido/zen/api_stream.rb
marksmith: This is how long the HTTP client may way for on read, before control returns, and the `APIStream` can detect the shutdown signal (`@running` has been set to `false`). There isn't a good way to interrupt the HTTP client, (at least not one that I could find).

lib/aikido/zen/agent.rb
timokoessler: The diff looked really strange, and Ruby is even worse ofc. 🙈

lib/aikido/zen/agent.rb
marksmith: Discussed internally and decided to leave this as is.

tomaisthorpe (commented):
I think not covered in this PR is the change of URL for zen realtime. We're moving to `https://zen.aikido.dev`, but as part of that, we need to ensure we can still fallback to the old URL incase there's network level blocking.

You can see how node is handling it here:
https://github.com/AikidoSec/firewall-node/blob/16a288a5880380ad540017397d56e3597b6d4a71/library/agent/realtime/probeRealtimeURL.ts

test/aikido/zen/runtime_settings_test.rb
hansott: Need to update this line?

hansott: also in more sinks: `"configUpdatedAt" => 1717171717000,`

test/support/wait_helpers.rb
```diff
@@ -0,0 +1,10 @@
+# frozen_string_literal: true
+
+module WaitHelpers
+  def wait_until(timeout:)
+    start_time = Time.now
+    until yield || (Time.now - start_time) > timeout
+      sleep 0.01
```
hansott, line 7: this is a blocking call? but since different thread no problem?

lib/aikido/zen/api_client.rb
```diff
@@ -16,6 +16,12 @@ def initialize(
       @config = config
       @system_info = system_info
       @rate_limiter = rate_limiter
+
+      @should_fetch_settings_endpoint = @config.realtime_endpoint
+    end
+
+    def should_fetch_settings_endpoint=(uri)
```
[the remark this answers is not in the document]
marksmith: This sets the endpoint used by `#should_fetch_settings?`.

test/support/wait_helpers.rb
marksmith: This method is currently used wait until connections arrive, from `Aikido::Zen::APIStream`, which does start and run in another thread. It was added to avoid busy-waiting.

test/aikido/zen/agent_test.rb
```diff
@@ -87,22 +111,98 @@ def report(event)
...
+    @agent.start!
+
+    assert_requested request
+
+    assert_logged :debug, /error probing realtime endpoint/i
+    refute_logged :error, /error probing realtime endpoint/i
+    assert_logged :warn, /can't reach #{Aikido::Zen.config.realtime_settings_updates_endpoint}/i
+  end
+
+  test "#start! probes the realtime endpont and logs warning after write timeout" do
```
timokoessler: Some more typos (not really important tho :D)

lib/aikido/zen/agent.rb
```diff
@@ -177,23 +197,68 @@ def send_heartbeat(at: Time.now.utc)
...
+
+    def heartbeats
       @heartbeats ||= Aikido::Zen::Agent::HeartbeatsManager.new(
         config: @config,
         worker: @worker
       )
     end
+
+    # @param data [Hash]
+    # @return [void]
```
timokoessler: It does not return void? At least it is used in a condition?

hansott (approved):

PR #286, opened 2026-05-27
Custom event tracking

marksmith:
This change adds support for tracking named user events, sent from application code using `Aikido::Zen.track_user_event`. User events include the event name, user ID, and IP address.

This change extends and should be reviewed after #285.

## Summary by Aikido
|  Security Issues: 0 | 🔍 Quality Issues: 4 |  Resolved Issues: 0 |
| :--- | :--- | :--- |

**🚀 New Features**
* Added user event tracking end-to-end with API and agent integration

**⚡ Enhancements**
* Handled empty API responses and adjusted event send endpoint

<sup>[More info](https://app.aikido.dev/featurebranch/scan/148343119?groupId=6)</sup>

### DIFF
--- a/docs/track.md
+++ b/docs/track.md
@@ -1,0 +1,55 @@
+# Tracking events
+
+`Aikido::Zen.track_user_event` lets you record things happening in your app — like failed logins, signups, or password resets. Zen sends these to Aikido so patterns can be detected, like someone failing to log in 50 times in a minute.
+
+```ruby
+# app/controllers/application_controller.rb
+class ApplicationController < ActionController::Base
+  private
+
+  def authenticate_user!
+    # Your authentication logic here
+    # ...
+
+    unless current_user
+        Aikido::Zen.track_user_event("user.login_failed")
+        return
+    end
+
+    Aikido::Zen.set_user(
+      id: current_user.id,
+      name: current_user.name
+    )
+
+    Aikido::Zen.track_user_event("user.login_succeeded")
+  end
+end
+```
+
+Zen automatically picks up the IP address, user agent, and current user (if you called [`setUser`](./user.md)) from the request — you don't need to pass those yourself.
+
+## More examples
+
+```ruby
+Aikido::Zen.track_user_event("user.signed_up")
+Aikido::Zen.track_user_event("user.password_reset_requested")
+Aikido::Zen.track_user_event("plan.invite_sent")
+Aikido::Zen.track_user_event("payment.failed")
+```
+
+## Naming events
+
+Use lowercase with dots to group related events:
+
+- `user.login_failed`
+- `user.login_succeeded`
+- `user.signed_up`
+- `user.password_reset_requested`
+- `payment.failed`
+- `plan.invite_sent`
+
+## Things to know
+
+`Aikido::Zen.track_user_event` only works inside an HTTP request. If you call it in a background job or a script, nothing gets sent and you'll see a warning in the console.
+
+If you haven't called `Aikido::Zen.set_user` yet, the event still goes through — it just won't have a user ID attached.
--- a/lib/aikido/zen.rb
+++ b/lib/aikido/zen.rb
@@ -216,6 +216,25 @@ module Aikido
       alias_method :set_user, :track_user
     end
 
+    # Track user event with name
+    #
+    # @param name [String]
+    # @return [void]
+    def self.track_user_event(name)
+      context = current_context
+      return unless context
+
+      request = context.request
+
+      event = Aikido::Zen::UserEvent.new(
+        name: name,
+        user_id: request.actor&.id,
+        ip_address: request.client_ip
+      )
+
+      agent.send_user_event(event)
+    end
+
     # @return [Aikido::Zen::AttackWave::Detector] the attack wave detector.
     def self.attack_wave_detector
       @attack_wave_detector ||= AttackWave::Detector.new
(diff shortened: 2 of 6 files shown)


lib/aikido/zen/agent.rb
```diff
@@ -177,23 +201,88 @@ def send_heartbeat(at: Time.now.utc)
...
+      # @yield the block to execute
+      # @yieldparam args [Array] the positional arguments passed to the method
+      # @yieldparam blk [Proc] the block passed to the method
+      # @yieldparam kwargs [Hash] the keyword arguments passed to the method
+      # @yieldreturn [Object] the return value of the method
+      # @return [void]
+      def updater(method_name, &block)
+        raise ArgumentError, "block required" unless block
+
+        instance_variable = :"@__updater_#{block.object_id}"
```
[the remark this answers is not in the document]
marksmith: The `block.object_id` was chosen to be safe and unique. Not all method names are valid instance variables, requiring normalization for safety.

docs/track.md
```diff
@@ -0,0 +1,55 @@
...
+# app/controllers/application_controller.rb
+class ApplicationController < ActionController::Base
+  private
+
+  def authenticate_user!
+    # Your authentication logic here
+    # ...
+
+    unless current_user
+        Aikido::Zen.track_user_event("user.login_failed")
```
timokoessler, line 15: Node.js only names the method `track`. I think we should keep it consistent. As the event can also be sent without a user, just for an ip, I would prefer the shorter one.

lib/aikido/zen.rb
```diff
@@ -216,6 +216,25 @@ class << self
       alias_method :set_user, :track_user
     end
 
+    # Track user event with name
+    #
+    # @param name [String]
+    # @return [void]
+    def self.track_user_event(name)
+      context = current_context
+      return unless context
```
timokoessler, line 225: In Node.js we log a warning (only once) if called without context. Otherwise user might never notice it is not working?

lib/aikido/zen/api_client.rb
```diff
@@ -117,6 +117,30 @@ def report(event)
...
+      event_type = "user_event"
+
+      if @rate_limiter.throttle?(event_type)
+        @config.logger.error("Not reporting #{event_type.upcase} event due to rate limiting")
+        return
+      end
+
+      @config.logger.debug("Reporting #{event_type.upcase} event")
+
+      req = Net::HTTP::Post.new("/api/runtime/events", default_headers)
```
timokoessler, line 130: These events can only be sent to `zen.aikido.dev`, the old runtime API does not support this.

lib/aikido/zen/api_client.rb
```diff
@@ -117,6 +117,30 @@ def report(event)
       raise
     end
 
+    def send_user_event(event)
+      event_type = "user_event"
+
+      if @rate_limiter.throttle?(event_type)
```
timokoessler, line 123: As these events are not send to Aikido Core, maybe not re-use the same rate limiter? (I think this is the case)

lib/aikido/zen/api_client.rb
tomaisthorpe: That's quite an important point for the other agents too. If the agent is unable to connect to `zen.aikido.dev`, we need to make it _really_ clear that the user events aren't going to work. I wonder if we need a log here, @timokoessler?

timokoessler: Indeed, makes sense. Maybe log if user calls the track() API for the first time? cc. @hansott.

docs/track.md
marksmith: The Ruby agent already uses `track_` internally in many cases. Adding a method named `track` is confusing and inconsistent in context. It would be nice if the agents agreed in all cased but Zen Ruby has done its own thing since the beginning.

We added an alias for Zen Node like `set_user` to `track_user`, but we did not anticipate Zen Node adding a method named `track`. If we are going to align long term then the alignment is going to have to go both ways, otherwise, Zen Node creates busywork for the other agents, where all other agents constantly need to adapt to what Zen Node does, as Zen Node freely goes against what the other agents do.

lib/aikido/zen.rb
marksmith: Zen Ruby always creates a context when handling requests. Outside of middleware run before our middleware, in practice it should not be possible for users to call `track_user_event` without a context. We could still log a warning in that case but it is unlikely anyone will see it.

lib/aikido/zen/event.rb
```diff
@@ -96,4 +96,20 @@ def as_json
       end
     end
   end
+
+  class UserEvent
```
timokoessler, line 100: Todo: This event needs to be updated to follow new spec:  https://github.com/AikidoSec/zen-specs/blob/main/features/custom_event_tracking.md

lib/aikido/zen/api_client.rb
marksmith: This is being considered as a future improvement.

lib/aikido/zen/agent.rb
```diff
@@ -165,6 +165,15 @@ def report(event)
       end
     end
 
+    # Reports a custom event tracked via Aikido::Zen.track_custom_event, if
+    # reporting is enabled.
+    #
+    # @param event [Aikido::Zen::Events::Custom]
+    # @return [void]
+    def report_custom_event(event)
+      report(event) if @api_client.can_make_requests?
```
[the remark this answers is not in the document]
marksmith: This is being considered as a future improvement.

hansott (approved):

PR #307, opened 2026-06-19
Reimplement multiprocess architecture

marksmith:
This change resolves several long-standing concerns with DRb, by replacing DRb for inter-process communication with an appropriate IPC and RPC mechanism; enabling the significant cleanup and simplification of the multiprocessor architecture.

The change begins by correcting another long-standing concern. `Hash` objects produced by `#as_json` and consumed by `#from_json` must have `String` keys. This is addressed first order to remove sources of errors, since the RPC mechanism uses JSON encoding.

The IPC mechanism uses an ephemeral port and an in-memory cryptographically secure random secret that are inherited by child processes at fork and used to restrict connection to real child processes using a mutual challenge-response protocol (built on HMAC-SHA256 and constant time comparison from OpenSSL).

The RPC mechanism is built on top of the IPC mechanism and is similar conceptually JSONRPC-2.0.

The multiprocessor architecture now consists of two parts.`Aikido::Zen::WorkerProcess::Agent::Server` runs in the parent process and is started at startup, and `Aikido::Zen::WorkerProcess::Agent::Client` runs in the child process and is started when a child process detected and handles the processes forking (on the first request after forking). The client in the child process makes requests of the server in the parent process. Two kinds of request are supported:

- synchronous requests, issued using invoke, which return a result
- asynchronous requests, issued using notify, which do not return a result

The parent processes no longer issues local requests using IPC. `Aikido::Zen::Agent` operates independently of the worker process server running in same processes or worker processes clients running in the child processes.

When the agent requests data from the API it caches the results. When the worker process requests data from the parent process no serialization is performed. The server returns the already-serialized cached data from the API and the child process handles it using the same logic as the parent processes.

## Summary by Aikido
| ⚠️ Security Issues: 13 | 🔍 Quality Issues: 8 |  Resolved Issues: 0 |
| :--- | :--- | :--- |

**🚀 New Features**
* Reimplemented multiprocess architecture using IPC/RPC for parent-child communication

**⚡ Enhancements**
* Added API cache and delegated rate limit calculation to worker client
* Updated CI to install mock server and run Rails end-to-end task

**🐛 Bugfixes**
* Enforced JSON hashes to use String keys in as_json/from_json

**🔧 Refactors**
* Replaced fork check with atomic PID compare-and-set in ForkDetector

<sup>[More info](https://app.aikido.dev/featurebranch/scan/147883174?groupId=6)</sup>


marksmith:
Tests for IPC and RPC to follow.

marksmith:
`#notify` has been removed, since it was only used in one place and that use case could be effectively replaced by responding early (before the handler has completed processing).

Tests have now been added for then IPC and RPC mechanisms. Removing `#notify` significantly reduced the number of tests that were required.

hansott (approved):

PR #313, opened 2026-06-25
Block home-relative paths

marksmith:
`File.expand_path` performs tilde expansion of home-relative paths after the paths are scanned to prevent path traversal attacks into absolute paths that may start with dangerous path starts. This change blocks paths that begin with tilde when the user input is contained in the file path presented for scanning, and so blocks home-relative paths.

## Summary by Aikido
|  Security Issues: 0 |  Quality Issues: 0 |  Resolved Issues: 0 |
| :--- | :--- | :--- |

**⚡ Enhancements**
* Blocked home-relative (~) paths and tightened path traversal predicate

<sup>[More info](https://app.aikido.dev/featurebranch/scan/137868510?groupId=6)</sup>

### DIFF
--- a/lib/aikido/zen/scanners/path_traversal_scanner.rb
+++ b/lib/aikido/zen/scanners/path_traversal_scanner.rb
@@ -45,6 +45,9 @@ module Aikido::Zen
       end
 
       def attack?
+        # We block home-relative path
+        return true if @filepath.start_with?("~") && @filepath.include?(@input)
+
         # Single character are ignored because they don't pose a big threat
         return false if @input.length <= 1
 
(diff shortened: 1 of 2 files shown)


lib/aikido/zen/scanners/path_traversal_scanner.rb
```diff
@@ -45,6 +45,9 @@ def initialize(filepath, input)
       end
 
       def attack?
+        # We block home-relative path
+        return true if @filepath.start_with?("~") && @filepath.include?(@input)
```
timokoessler: I guess the user input also needs to contain the tilde character?

timokoessler (approved):

PR #343, opened 2026-07-29
Support gradual rollout flag for realtime updates

marksmith:
This change exposes the new `realtimeUpdatesEnabled` flag from the runtime config response as `Aikido::Zen::RuntimeSettings#realtime_settings_updates_enabled`, and only starts the API stream once the first settings update has been applied and either `Aikido::Zen::RuntimeSettings#realtime_settings_updates_enabled`  or `Aikido::Zen::Config#realtime_settings_updates_enabled` is truthy.

### DIFF
--- a/lib/aikido/zen/agent.rb
+++ b/lib/aikido/zen/agent.rb
@@ -71,7 +71,7 @@ module Aikido::Zen
         if update_settings_from_runtime_config!(response, reason: "after start")
           updated_settings!
 
-          if @config.realtime_settings_updates_enabled? || @settings.realtime_settings_updates_enabled
+          if @config.realtime_settings_updates_enabled? || @settings.realtime_settings_updates_enabled?
             @api_stream.handle("config-updated") do |event|
               @config.logger.debug("Received server-sent event: config-updated")
               settings_updated(event)
--- a/lib/aikido/zen/agent.rb
+++ b/lib/aikido/zen/agent.rb
@@ -18,12 +18,14 @@ module Aikido::Zen
 
     def initialize(
       config: Aikido::Zen.config,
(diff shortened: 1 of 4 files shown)


lib/aikido/zen/runtime_settings.rb
```diff
@@ -97,6 +121,8 @@ def update_from_runtime_config_json(data)
 
       self.excluded_user_ids_from_rate_limiting = data["excludedUserIdsFromRateLimiting"]
 
+      self.enabled_features = Set.new(data["enabledFeatures"])
```
timokoessler, line 124: This is safe it not returned by server?

lib/aikido/zen/agent.rb
```diff
@@ -197,7 +199,7 @@ def send_heartbeat(at: Time.now.utc)
     # @return [void]
     # @see Aikido::Zen::RuntimeSettings
     def poll_for_setting_updates
-      @worker.every(@config.polling_interval) do
+      @worker.every(@config.polling_interval, run_now: true) do
```
timokoessler, line 202: This is changed because of the refactor above?

lib/aikido/zen/runtime_settings.rb
marksmith: Yes. If the key does not exist, `data["enabledFeatures"]` will return `nil`, and `Set.new` return an empty `Set`.

lib/aikido/zen/agent.rb
marksmith: This change isn't strictly necessary, but it makes the default explicit at the call site; the block that is passed to `@worker.every` here was already being `run_now`.

timokoessler (approved):

PR #369, opened 2026-08-11
Improve attack wave support for multi-process

hansott:
When running with multiple puma workers, the attack wave suspicious request count is stored per worker. Similarly to reporting heartbeats with the request stats, we should count stats across workers and merge them. This ensures that 15 reqs to multiple workers triggers one attack wave.


marksmith:
Superseded by #372.

PR #372, opened 2026-08-13
Centralize attack wave detection state

marksmith:
This change centralizes attack wave detection state in the main process, and provides access to this centralized state in worker processes through the `record_attack_wave` and `attack_wave_samples` RPC calls. Should an RPC call fail then the worker process falls back to the local attack wave detection state until the next RPC call. For sustained PRC failure this gracefully moves attack wave detection state to the worker process.

This helps ensure that an attack waves is detected after the configured attack wave threshold and are tracked once, in the process that detected it.

The attack wave detection state is now separate from the attack wave detection logic itself. Attack wave detection is still performed in the worker process. Attack wave detection state is centralized in the main process under normal operation. `Aikido::Zen` orchestrates attack wave detection so that it works correctly in the main process and worker processes.

The pre-existing optimization of whether the client IP has been flagged as the origin of an attack wave has been retired due to the complexity and cost of synchronizing the flagged cache state. `Aikido::Zen::AttackWave::Detector#record` now does this check only to guarantee the minimum time between recording attack wave events.

This change supersedes #369.

### DIFF
--- a/lib/aikido/zen.rb
+++ b/lib/aikido/zen.rb
@@ -268,6 +268,13 @@ module Aikido
     # triggering an attack wave is crossed. If the threshold is crossed,
     # the client IP is flagged as having just triggered an attack wave.
     #
+    # In multiprocess deployments, on RPC failure, the worker process
+    # records against its local detector.
+    #
+    # If RPC failures are intermittent, it is possible that an attack
+    # wave may be missed or duplicated, because state is split between
+    # the global and local detectors.
+    #
     # @param client_ip [String]
     # @param sample [Aikido::Zen::AttackWave::Sample]
     # @return [Boolean]
--- a/lib/aikido/zen.rb
+++ b/lib/aikido/zen.rb
@@ -249,24 +249,24 @@ module Aikido
       @attack_wave_detector ||= AttackWave::Detector.new
     end
 
-    # Classifies the current request and returns whether the threshold for
-    # triggering an attack wave is crossed.
+    # Classifies the current request and, if it is classified as suspicious,
+    # records a sample against the attack wave detector.
     #
     # @param context [Aikido::Zen::Context]
     # @param status_code [Integer, nil]
-    # @return [Boolean]
-    def self.attack_wave?(context, status_code = nil)
+    # @return [Array<Aikido::Zen::AttackWave::Sample>, nil]
+    def self.detect_attack_wave(context, status_code = nil)
       client_ip = context.request.client_ip
-      return false unless client_ip
+      return nil unless client_ip
 
-      return false unless AttackWave::Helpers.web_scanner?(context, status_code)
+      return nil unless AttackWave::Helpers.web_scanner?(context, status_code)
 
       record_attack_wave(client_ip, AttackWave::Helpers.sample_for(context))
     end
 
-    # Records a suspicious sample and returns whether the threshold for
-    # triggering an attack wave is crossed. If the threshold is crossed,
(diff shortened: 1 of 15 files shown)


lib/aikido/zen/attack_wave.rb
```diff
@@ -5,13 +5,20 @@
 
 module Aikido::Zen
   module AttackWave
+    # Tracks per-client-IP attack wave state.
+    #
```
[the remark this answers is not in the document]
marksmith: Documents intention.

lib/aikido/zen/worker_process/agent/server.rb
```diff
@@ -25,6 +26,14 @@ def initialize(config: Aikido::Zen.config)
           result = calculate_rate_limits(route_data, ip, actor_data)
           respond.call(result&.as_json)
         end
+
+        @rpc_server.handle("record_attack_wave") do |respond, client_ip, verb, path|
+          respond.call(record_attack_wave(client_ip, verb, path))
+        end
+
+        @rpc_server.handle("attack_wave_samples") do |respond, client_ip|
+          respond.call(attack_wave_samples(client_ip))
```
timokoessler: Record uses `@mutex.synchronize`, do we also need to check mutex when reading samples?

lib/aikido/zen/worker_process/agent/server.rb
```diff
@@ -92,6 +100,17 @@ def calculate_rate_limits(route_data, ip, actor_data)
...
+
+      # @return [Boolean]
+      def record_attack_wave(client_ip, verb, path)
+        sample = Aikido::Zen::AttackWave::Sample.new(verb: verb, path: path)
+        @detector.record(client_ip, sample)
+      end
+
+      # @return [Array<Hash>]
+      def attack_wave_samples(client_ip)
+        @detector.samples[client_ip].to_a.map(&:as_json)
```
timokoessler: Also no mutex needed here for reading, as @detector.record is setting one when writing?

lib/aikido/zen/worker_process/agent/server.rb
marksmith: Strictly, no, because once the attack wave is detected no new samples are added.

lib/aikido/zen.rb
```diff
@@ -249,6 +249,61 @@ def self.attack_wave_detector
...
+    # @param sample [Aikido::Zen::AttackWave::Sample]
+    # @return [Boolean]
+    def self.record_attack_wave(client_ip, sample)
+      worker_process_client = @worker_process_client
+
+      if worker_process_client
+        begin
+          worker_process_client.record_attack_wave(client_ip, sample)
+        rescue
+          attack_wave_detector.record(client_ip, sample)
```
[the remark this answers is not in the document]
marksmith: This is a known, unavoidable issue. The fallback to the local attack wave detector is a best-effort attempt to keep detection mostly functional in the worst case. This is now noted as a comment.

lib/aikido/zen/worker_process/agent/server.rb
marksmith: `attack_wave_samples` has been removed due to an RPC failure edge case.

lib/aikido/zen/worker_process/agent/server.rb
marksmith: `attack_wave_samples` has been removed due to an RPC failure edge case.

timokoessler (approved):

PR #386, opened 2026-08-25
Fix SSRF OAuth iss query parameter false positive

marksmith:
This change adds the `ignore_iss_query_parameter` configuration option to control whether the `iss` query parameter is checked by the SSRF scanner. This configuration option defaults to `false`, because not all applications use OAuth/OIDC and not all deployments are affected by this false positive:

RFC 9207 mandates that OAuth/OIDC providers echo their own URI back in the `iss` query parameter on the callback URI, for the client to validate against. This causes `Aikido::Zen::Scanners::SSRFScanner` to wrongly flag the `iss` parameter as an SSRF attack when the provider is has a private IP address, because the `iss` value matches the hostname that your application will connect to for configuration or token exchange.

If `Aikido::Zen.ignore_iss_query_parameter` is set, the SSRF scanner ignores the `iss` query parameter for every request, not just the OAuth/OIDC callback.


marksmith:
This PR is being closed because it is not immediately required and we are deciding how best to handle false positives.

PR #387, opened 2026-08-27
Add common container/cloud directories

marksmith:
This change adds four reported widely-deployed Docker/container-specific unsafe path starts:

- /data/
- /rails/
- /workspace/
- /workspaces/


lib/aikido/zen/scanners/path_traversal/helpers.rb
```diff
@@ -27,7 +27,11 @@ module PathTraversal
         "/var/",
         # Common container/cloud directories
         "/app/",
-        "/code/"
+        "/code/",
+        "/data/",
```
tomaisthorpe, line 31: `/data/` might be too generic? I've definitely used it before in a docker mount.

timokoessler (approved):

